PR #71, opened 2021-01-09
enable to install under proxy

yamato225:
To install TinyPilot using proxy, I added "extra_args" on "create TinyPilot virtualenv" step.

When I tried to install using proxy, this step failed because pip command can't verify SSL certification.
I could solve this problem by adding "--trusted-host pypi.org" arguments on this step.

But it's not good for the user who doesn't use proxy, so I use `pip_args` variable.
If the user want to use proxy, `pip_args` is "--trusted-host pypi.org".
If not, `pip_args` is empty text.

pip_args is setten by `quick-install`.

Related: https://github.com/mtlynch/tinypilot/pull/402


yamato225:
test failed because `pip_args` is not defined. I close this PR.

PR #72, opened 2021-01-09
Enable install under proxy

yamato225:
To install TinyPilot using proxy, I added "extra_args" on "create TinyPilot virtualenv" step.

When I tried to install using proxy, this step failed because pip command can't verify SSL certification.
I could solve this problem by adding "--trusted-host pypi.org" arguments on this step.

But it's not good for the user who doesn't use proxy, so I use pip_args variable.
If the user want to use proxy, pip_args is "--trusted-host pypi.org".
If not, `pip_args` is empty text.

pip_args is setten by quick-install.

Related: https://github.com/mtlynch/tinypilot/pull/402

### DIFF
--- a/defaults/main.yml
+++ b/defaults/main.yml
@@ -15,4 +15,4 @@ tinypilot_port: 8000
 tinypilot_keyboard_interface: /dev/hidg0
 tinypilot_mouse_interface: /dev/hidg1
 tinypilot_enable_debug_logging: no
-pip_args: ""
+tinypilot_pip_args: ""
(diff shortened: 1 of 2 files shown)


defaults/main.yml
```diff
@@ -15,3 +15,4 @@ tinypilot_port: 8000
 tinypilot_keyboard_interface: /dev/hidg0
 tinypilot_mouse_interface: /dev/hidg1
 tinypilot_enable_debug_logging: no
+pip_args: ""
```
mtlynch, line 18: Can we make this `tinypilot_pip_args` for consistency and leave a trailing newline?

mtlynch (changes requested):

yamato225:
Sure. I changed `pip_args` to `tinypilot_pip_args`.

mtlynch:
Thanks!

PR #80, opened 2021-01-22
Add collect-debug-logs to tinypilot-privileged directory

mtlynch:
Related: https://github.com/mtlynch/tinypilot/issues/426

### DIFF
--- a/files/collect-debug-logs
+++ b/files/collect-debug-logs
@@ -1,0 +1,78 @@
+#!/bin/bash
+
+# Exit on unset variable.
+set -u
+
+LOG_FILE=$(mktemp)
+echo "Writing diagnostic logs to $LOG_FILE"
+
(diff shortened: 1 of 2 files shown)


jotaen:
@mtlynch Is there a reson why `collect-debug-logs` has different permissions than the other scripts?

```
-rw-r--r--   collect-debug-logs
-rwxr-xr-x   init-usb-gadget
-rwxr-xr-x   update
```

mtlynch:
Good question! No, just a mistake. #91 fixed it.

jotaen:
Also I believe `[-hsu]` in the help text might not be correct?

mtlynch:
>Also I believe [-hsu] in the help text might not be correct?

Oh, good catch! Fixed in #93.

PR #94, opened 2021-02-12
Add script to change machine’s hostname

jotaen:
This new bash script changes the hostname of the machine, so it populates a new hostname to `/etc/hostname` and `/etc/hosts` as discussed in https://github.com/mtlynch/tinypilot/issues/433.

Just to double-check: while I think technically possible, I don’t think the hostname should be allowed to contain any dots, so I didn’t allow this.

---
This change is [Reviewable](https://reviewable.io/reviews/mtlynch/ansible-role-tinypilot/94)


mtlynch (changes requested):
> *[Reviewable](https://reviewable.io/reviews/mtlynch/ansible-role-tinypilot/94)* status: 0 of 2 files reviewed, 4 unresolved discussions (waiting on @jotaen)

___
*[files/change-hostname, line 16 at r1](https://reviewable.io/reviews/mtlynch/ansible-role-tinypilot/94#-MTN3T843bOAgxdlpeBB:-MTN3T843bOAgxdlpeBC:blponzs) ([raw file](https://github.com/mtlynch/ansible-role-tinypilot/blob/27b8fc75cf76d30d0b384997306cd2dfdc249546/files/change-hostname#L16)):*
> ```
> # Parse command-line arguments.
> while getopts "h" opt; do
>   case $opt in
> ```

I've been bad at this, but I'm trying to ensure new code adheres to the Google shell style guidelines that say to quote and curly brace variable names:

https://google.github.io/styleguide/shellguide.html#variable-expansion

(ditto throughout)
___
*[files/change-hostname, line 30 at r1](https://reviewable.io/reviews/mtlynch/ansible-role-tinypilot/94#-MTN32GC-u0WWBNMvvtE:-MTN32GC-u0WWBNMvvtF:b-k39dzv) ([raw file](https://github.com/mtlynch/ansible-role-tinypilot/blob/27b8fc75cf76d30d0b384997306cd2dfdc249546/files/change-hostname#L30)):*
> ```
> shift $(($OPTIND - 1))
> if [[ ! $@ ]]; then
> 	echo "Input parameter missing: new_hostname"
> ```

Let's keep consistently to spaces:

https://google.github.io/styleguide/shellguide.html#indentation

We should also print to stderr rather than stdout:

https://google.github.io/styleguide/shellguide.html#stdout-vs-stderr

(ditto throughout)
___
*[files/change-hostname, line 52 at r1](https://reviewable.io/reviews/mtlynch/ansible-role-tinypilot/94#-MTN3qHmA8OAB87rAg2J:-MTN3qHmA8OAB87rAg2K:b-koh7bp) ([raw file](https://github.com/mtlynch/ansible-role-tinypilot/blob/27b8fc75cf76d30d0b384997306cd2dfdc249546/files/change-hostname#L52)):*
> ```
> 
> # Retrieve original `/etc/hosts`.
> # - Read file line by line, make sure to preserve
> ```

We're line breaking a bit early here. Can we go until 80 characters?

https://google.github.io/styleguide/shellguide.html#line-length-and-long-strings
___
*[files/change-hostname, line 76 at r1](https://reviewable.io/reviews/mtlynch/ansible-role-tinypilot/94#-MTN5hhq2SBboPYWUJlV:-MTN5hhr6f7SKTgezQHk:b-lzon55) ([raw file](https://github.com/mtlynch/ansible-role-tinypilot/blob/27b8fc75cf76d30d0b384997306cd2dfdc249546/files/change-hostname#L76)):*
> ```
> # Populate new hostname to system files.
> OLD_ETC_HOSTS=$(printf "%s\n" "${ETC_HOSTS_ORIGINAL[@]}")
> NEW_MARKER_SECTION="$MARKER_START\n127.0.0.1 $NEW_HOSTNAME\n$MARKER_END"
> ```

Debian (and so Raspbian too) uses `127.0.1.1` rather than `127.0.0.1`for the local hostname, oddly:

https://www.debian.org/doc/manuals/debian-reference/ch05.en.html#_the_hostname_resolution

mtlynch (changes requested):
> *[Reviewable](https://reviewable.io/reviews/mtlynch/ansible-role-tinypilot/94)* status: 0 of 2 files reviewed, 4 unresolved discussions (waiting on @jotaen)

___
*[files/change-hostname, line 76 at r1](https://reviewable.io/reviews/mtlynch/ansible-role-tinypilot/94#-MTN5hhq2SBboPYWUJlV:-MTN87SUEgNOiPnMocci:b-kaj9x) ([raw file](https://github.com/mtlynch/ansible-role-tinypilot/blob/27b8fc75cf76d30d0b384997306cd2dfdc249546/files/change-hostname#L76)):*

Debian (and so Raspbian too) uses `127.0.1.1` rather than `127.0.0.1`for the local hostname, oddly:

https://www.debian.org/doc/manuals/debian-reference/ch05.en.html#_the_hostname_resolution
</blockquote></details>

Oh, this is on me. I wrote it wrong in the bug originally.

jotaen:
Would you be open to adding an [`.editorconfig`](https://editorconfig.org/) regarding the tabs/spaces thing? Happy to submit a PR.

jotaen (commented):
> *[Reviewable](https://reviewable.io/reviews/mtlynch/ansible-role-tinypilot/94)* status: 0 of 2 files reviewed, 4 unresolved discussions (waiting on @mtlynch)

___
*[files/change-hostname, line 16 at r1](https://reviewable.io/reviews/mtlynch/ansible-role-tinypilot/94#-MTN3T843bOAgxdlpeBB:-MTNCO8d4TcmXBi36q3U:b-qba3ps) ([raw file](https://github.com/mtlynch/ansible-role-tinypilot/blob/27b8fc75cf76d30d0b384997306cd2dfdc249546/files/change-hostname#L16)):*

I've been bad at this, but I'm trying to ensure new code adheres to the Google shell style guidelines that say to quote and curly brace variable names:

https://google.github.io/styleguide/shellguide.html#variable-expansion

(ditto throughout)
</blockquote></details>

Will need to take a closer look at the shell script style guide, was actually not sure about that because I saw different styles being used in the other scripts. I’ll make sure to follow it.
___
*[files/change-hostname, line 30 at r1](https://reviewable.io/reviews/mtlynch/ansible-role-tinypilot/94#-MTN32GC-u0WWBNMvvtE:-MTNCFWA9SpVa6ZoZ-ut:b-896fix) ([raw file](https://github.com/mtlynch/ansible-role-tinypilot/blob/27b8fc75cf76d30d0b384997306cd2dfdc249546/files/change-hostname#L30)):*

Let's keep consistently to spaces:

https://google.github.io/styleguide/shellguide.html#indentation

We should also print to stderr rather than stdout:

https://google.github.io/styleguide/shellguide.html#stdout-vs-stderr

(ditto throughout)
</blockquote></details>

Done.
___
*[files/change-hostname, line 52 at r1](https://reviewable.io/reviews/mtlynch/ansible-role-tinypilot/94#-MTN3qHmA8OAB87rAg2J:-MTNCLNgCdqJjqLv9MDn:b-896fix) ([raw file](https://github.com/mtlynch/ansible-role-tinypilot/blob/27b8fc75cf76d30d0b384997306cd2dfdc249546/files/change-hostname#L52)):*

We're line breaking a bit early here. Can we go until 80 characters?

https://google.github.io/styleguide/shellguide.html#line-length-and-long-strings
</blockquote></details>

Done.
___
*[files/change-hostname, line 76 at r1](https://reviewable.io/reviews/mtlynch/ansible-role-tinypilot/94#-MTN5hhq2SBboPYWUJlV:-MTNCMKGAXAF8Wdam1f_:b-896fix) ([raw file](https://github.com/mtlynch/ansible-role-tinypilot/blob/27b8fc75cf76d30d0b384997306cd2dfdc249546/files/change-hostname#L76)):*

Oh, this is on me. I wrote it wrong in the bug originally.
</blockquote></details>

Done.

mtlynch:
>Would you be open to adding an .editorconfig regarding the tabs/spaces thing? Happy to submit a PR.

Sure, that sounds good to me.

We could also add a CI check for it like [check-trailing-newline](https://github.com/mtlynch/ansible-role-tinypilot/blob/0e65db74753df02eb77c181efd93aa91d8f4ef33/tests/check-trailing-newline).

mtlynch (changes requested):
> *[Reviewable](https://reviewable.io/reviews/mtlynch/ansible-role-tinypilot/94)* status: 0 of 2 files reviewed, 2 unresolved discussions (waiting on @jotaen)

___
*[files/change-hostname, line 16 at r1](https://reviewable.io/reviews/mtlynch/ansible-role-tinypilot/94#-MTN3T843bOAgxdlpeBB:-MTNDWZK8TjV2rSYlLbH:b-3e2v5) ([raw file](https://github.com/mtlynch/ansible-role-tinypilot/blob/27b8fc75cf76d30d0b384997306cd2dfdc249546/files/change-hostname#L16)):*
<details><summary><i>Previously, jotaen (Jan Heuermann) wrote…</i></summary><blockquote>

Will need to take a closer look at the shell script style guide, was actually not sure about that because I saw different styles being used in the other scripts. I’ll make sure to follow it.
</blockquote></details>

Oh, to clarify, can we rewrite the variable names here in Google style?

So:

* `$opt` -> `"${opt}"`
* `"$NEW_HOSTNAME"` -> `"${NEW_HOSTNAME}"`
* etc.

Also, I just realized from reviewing the Google style doc that I've been doing bash variables mostly wrong up to this point. 🙃 Apparently, uppercase is only for constants (and we should be explicitly declaring constants with `readonly`).
___
*[files/change-hostname, line 30 at r2](https://reviewable.io/reviews/mtlynch/ansible-role-tinypilot/94#-MTNDOvi56daZdpD0Fbt:-MTNDOvi56daZdpD0Fbu:byd7hjq) ([raw file](https://github.com/mtlynch/ansible-role-tinypilot/blob/40af4c99a5434097a114fd70100ad5db29fe29d7/files/change-hostname#L30)):*
> ```
> shift $(($OPTIND - 1))
> if [[ ! $@ ]]; then
>   echo "Input parameter missing: new_hostname" 1>&2
> ```

I'm not familiar with this syntax. Why `1>&2` as opposed to `>&2` without the `1`?

jotaen (commented):
> *[Reviewable](https://reviewable.io/reviews/mtlynch/ansible-role-tinypilot/94)* status: 0 of 2 files reviewed, 2 unresolved discussions (waiting on @jotaen and @mtlynch)

___
*[files/change-hostname, line 30 at r2](https://reviewable.io/reviews/mtlynch/ansible-role-tinypilot/94#-MTNDOvi56daZdpD0Fbt:-MTNGer7EUyp2NF4LCPd:b-ijfoqe) ([raw file](https://github.com/mtlynch/ansible-role-tinypilot/blob/40af4c99a5434097a114fd70100ad5db29fe29d7/files/change-hostname#L30)):*

I'm not familiar with this syntax. Why `1>&2` as opposed to `>&2` without the `1`?
</blockquote></details>

`>&2` is shorthand for `1>&2`. I personally use the latter out of habit, but I really should read that style guide first now :sweat_smile:

jotaen (commented):
> *[Reviewable](https://reviewable.io/reviews/mtlynch/ansible-role-tinypilot/94)* status: 0 of 2 files reviewed, 2 unresolved discussions (waiting on @mtlynch)

___
*[files/change-hostname, line 16 at r1](https://reviewable.io/reviews/mtlynch/ansible-role-tinypilot/94#-MTN3T843bOAgxdlpeBB:-MTW3XjB3bMFjT_rY7L4:b-896fix) ([raw file](https://github.com/mtlynch/ansible-role-tinypilot/blob/27b8fc75cf76d30d0b384997306cd2dfdc249546/files/change-hostname#L16)):*

Oh, to clarify, can we rewrite the variable names here in Google style?

So:

* `$opt` -> `"${opt}"`
* `"$NEW_HOSTNAME"` -> `"${NEW_HOSTNAME}"`
* etc.

Also, I just realized from reviewing the Google style doc that I've been doing bash variables mostly wrong up to this point. 🙃 Apparently, uppercase is only for constants (and we should be explicitly declaring constants with `readonly`).
</blockquote></details>

Done.
___
*[files/change-hostname, line 30 at r2](https://reviewable.io/reviews/mtlynch/ansible-role-tinypilot/94#-MTNDOvi56daZdpD0Fbt:-MTW3ZWpEb4oDsxFNiBK:b-z50boj) ([raw file](https://github.com/mtlynch/ansible-role-tinypilot/blob/40af4c99a5434097a114fd70100ad5db29fe29d7/files/change-hostname#L30)):*
<details><summary><i>Previously, jotaen (Jan Heuermann) wrote…</i></summary><blockquote>

`>&2` is shorthand for `1>&2`. I personally use the latter out of habit, but I really should read that style guide first now :sweat_smile: 
</blockquote></details>

Done. (Changed to `>&2`)

jotaen:
I revised the code style of the script according to [Google’s shell style guide](https://google.github.io/styleguide/shellguide.html) and it now also passes [shellcheck](https://www.shellcheck.net/).

One remark: I personally wouldn’t write comments such as `# Exit on unset variable.` for `set -u`, because for me this doesn’t add enough information over the code itself. In case you find `set -u` too undescriptive, there is also the long form (`set -o errexit`, `set -o nounset`, etc.) which I find pretty much self-documenting. I left it like this for now anyways (and even added `set -e`), because it’s done that way in the other scripts too, so I’d value the consistency higher for this PR.

Btw., maybe it would be good to revise all scripts in a separate refactoring PR to make them all nice and consisent. (Also the file names for example, which should use underscores instead of dashes.)

mtlynch (changes requested):
> *[Reviewable](https://reviewable.io/reviews/mtlynch/ansible-role-tinypilot/94)* status: 0 of 2 files reviewed, 5 unresolved discussions (waiting on @jotaen)

___
*[files/change-hostname, line 38 at r4](https://reviewable.io/reviews/mtlynch/ansible-role-tinypilot/94#-MTa0bl81nOa6YDKJq0F:-MTa0bl94DZxBr-wBSah:bn8y0rk) ([raw file](https://github.com/mtlynch/ansible-role-tinypilot/blob/8cca6293647ff7b14263a5aafff3f3e57b4af114/files/change-hostname#L38)):*
Why the switch from `$@`? It seems like the style guide prefers `"$@"` (with quotes):

>Use `"$@"` unless you have a specific reason to use `$*`, such as simply appending the arguments to a string in a message or log.

https://google.github.io/styleguide/shellguide.html#quoting
___
*[files/change-hostname, line 46 at r4](https://reviewable.io/reviews/mtlynch/ansible-role-tinypilot/94#-MTa0rwq1lQrQ_78iJUU:-MTa0rwq1lQrQ_78iJUV:b-t8wmse) ([raw file](https://github.com/mtlynch/ansible-role-tinypilot/blob/8cca6293647ff7b14263a5aafff3f3e57b4af114/files/change-hostname#L46)):*
> ```
> # See here for the rules:
> # https://man7.org/linux/man-pages/man7/hostname.7.html
> new_hostname="$1"
> ```

I think we want this to be:

```bash
readonly NEW_HOSTNAME="${1}"
```
___
*[files/change-hostname, line 47 at r4](https://reviewable.io/reviews/mtlynch/ansible-role-tinypilot/94#-MTa1gGaEsFJu7uLoXBq:-MTa1gGb-PqtkeoMQzoW:b-emi6ds) ([raw file](https://github.com/mtlynch/ansible-role-tinypilot/blob/8cca6293647ff7b14263a5aafff3f3e57b4af114/files/change-hostname#L47)):*
I'm possibly not communicating this correctly, or maybe we have a different understanding of the style guide.

In the thread about quoting, I said this:

> I'm trying to ensure new code adheres to the Google shell style guidelines that say to quote and curly brace variable names:

Based on my reading of the style guide, we should always be quoting and curly-bracing variable names, so the above code line should be written as:

`"${new_hostname} =~ "${VALID_HOSTNAME_PATTERN}"`

https://google.github.io/styleguide/shellguide.html#s5.7-quoting

This applies to variables everywhere in the script, so like:

`$is_in_marker_section || [[ "$line" == "$MARKER_START"`

I think is supposed to be:

`"${is_in_marker_section}" || [[ "${line}" == "${MARKER_START}"`

Is that your understanding?

(ditto throughout)
___
*[files/change-hostname, line 60 at r4](https://reviewable.io/reviews/mtlynch/ansible-role-tinypilot/94#-MTa2Hlm3IbPjompuqfK:-MTa2Hlm3IbPjompuqfL:b-cz4tlt) ([raw file](https://github.com/mtlynch/ansible-role-tinypilot/blob/8cca6293647ff7b14263a5aafff3f3e57b4af114/files/change-hostname#L60)):*
The style guide recommends quoting words:

>```bash
># "prefer quoting words", not compulsory
>readonly USE_INTEGER='true'
>```

https://google.github.io/styleguide/shellguide.html#s5.7-quoting

(ditto on lines 63 and 67)
___
*[files/change-hostname, line 79 at r4](https://reviewable.io/reviews/mtlynch/ansible-role-tinypilot/94#-MTa1FZF1SYu7i9FB4m_:-MTa1FZF1SYu7i9FB4ma:b-m5kd9a) ([raw file](https://github.com/mtlynch/ansible-role-tinypilot/blob/8cca6293647ff7b14263a5aafff3f3e57b4af114/files/change-hostname#L79)):*
> ```
> 
> # Populate new hostname to system files.
> old_etc_hosts=$(printf "%s\n" "${etc_hosts_original[@]}")
> ```

Can't this be `readonly` as well?

jotaen (commented):
> *[Reviewable](https://reviewable.io/reviews/mtlynch/ansible-role-tinypilot/94)* status: 0 of 2 files reviewed, 5 unresolved discussions (waiting on @mtlynch)

___
*[files/change-hostname, line 38 at r4](https://reviewable.io/reviews/mtlynch/ansible-role-tinypilot/94#-MTa0bl81nOa6YDKJq0F:-MTaAE3BA0jb-MPgip-y:b-2jvm4n) ([raw file](https://github.com/mtlynch/ansible-role-tinypilot/blob/8cca6293647ff7b14263a5aafff3f3e57b4af114/files/change-hostname#L38)):*

Why the switch from `$@`? It seems like the style guide prefers `"$@"` (with quotes):

>Use `"$@"` unless you have a specific reason to use `$*`, such as simply appending the arguments to a string in a message or log.

https://google.github.io/styleguide/shellguide.html#quoting
</blockquote></details>

It came recommended by shellcheck and I took it over without double-check the style guide again. Will change!
___
*[files/change-hostname, line 46 at r4](https://reviewable.io/reviews/mtlynch/ansible-role-tinypilot/94#-MTa0rwq1lQrQ_78iJUU:-MTaAap8BQUVE1MSEtLq:b-cv8p2b) ([raw file](https://github.com/mtlynch/ansible-role-tinypilot/blob/8cca6293647ff7b14263a5aafff3f3e57b4af114/files/change-hostname#L46)):*

I think we want this to be:

```bash
readonly NEW_HOSTNAME="${1}"
```
</blockquote></details>

The style guide says to NOT curly-brace default variables like `"$1"` unless you have a special case like `"${1}2"`

https://google.github.io/styleguide/shellguide.html#variable-expansion

You are right about the `readonly` and the all-caps of course.
___
*[files/change-hostname, line 47 at r4](https://reviewable.io/reviews/mtlynch/ansible-role-tinypilot/94#-MTa1gGaEsFJu7uLoXBq:-MTaAruC9gpUe8OZxkA0:b-fc54jh) ([raw file](https://github.com/mtlynch/ansible-role-tinypilot/blob/8cca6293647ff7b14263a5aafff3f3e57b4af114/files/change-hostname#L47)):*

I'm possibly not communicating this correctly, or maybe we have a different understanding of the style guide.

In the thread about quoting, I said this:

> I'm trying to ensure new code adheres to the Google shell style guidelines that say to quote and curly brace variable names:

Based on my reading of the style guide, we should always be quoting and curly-bracing variable names, so the above code line should be written as:

`"${new_hostname} =~ "${VALID_HOSTNAME_PATTERN}"`

https://google.github.io/styleguide/shellguide.html#s5.7-quoting

This applies to variables everywhere in the script, so like:

`$is_in_marker_section || [[ "$line" == "$MARKER_START"`

I think is supposed to be:

`"${is_in_marker_section}" || [[ "${line}" == "${MARKER_START}"`

Is that your understanding?

(ditto throughout)
</blockquote></details>

Starting with bash 3.2 there is a functional difference between `[[ "$foo" =~ "$pattern" ]]` (string matching) and `[[ "$foo" =~ $pattern ]]` (regex matching). I think the style guide doesn’t cover that case.

https://stackoverflow.com/questions/218156/bash-regex-with-quotes

Sorry about the `$is_in_marker_section` – up until just now I thought there exist boolean literals in bash… :facepalm: My bad! (ditto below for the assignment)
___
*[files/change-hostname, line 60 at r4](https://reviewable.io/reviews/mtlynch/ansible-role-tinypilot/94#-MTa2Hlm3IbPjompuqfK:-MTaCG4FBVyUh1A8LfR9:bq608yt) ([raw file](https://github.com/mtlynch/ansible-role-tinypilot/blob/8cca6293647ff7b14263a5aafff3f3e57b4af114/files/change-hostname#L60)):*

The style guide recommends quoting words:

>```bash
># "prefer quoting words", not compulsory
>readonly USE_INTEGER='true'
>```

https://google.github.io/styleguide/shellguide.html#s5.7-quoting

(ditto on lines 63 and 67)
</blockquote></details>

:+1: (see comment about boolean literals)
___
*[files/change-hostname, line 79 at r4](https://reviewable.io/reviews/mtlynch/ansible-role-tinypilot/94#-MTa1FZF1SYu7i9FB4m_:-MTaBlzXB9MnocS4gXTn:b4c9lyr) ([raw file](https://github.com/mtlynch/ansible-role-tinypilot/blob/8cca6293647ff7b14263a5aafff3f3e57b4af114/files/change-hostname#L79)):*

Can't this be `readonly` as well?
</blockquote></details>

:thumbsup:

jotaen (commented):
> *[Reviewable](https://reviewable.io/reviews/mtlynch/ansible-role-tinypilot/94)* status: 0 of 2 files reviewed, 5 unresolved discussions (waiting on @mtlynch)

___
*[files/change-hostname, line 38 at r4](https://reviewable.io/reviews/mtlynch/ansible-role-tinypilot/94#-MTa0bl81nOa6YDKJq0F:-MTkaDSXCj8lGPDlQpUe:brh2u3o) ([raw file](https://github.com/mtlynch/ansible-role-tinypilot/blob/8cca6293647ff7b14263a5aafff3f3e57b4af114/files/change-hostname#L38)):*
<details><summary><i>Previously, jotaen (Jan Heuermann) wrote…</i></summary><blockquote>

It came recommended by shellcheck and I took it over without double-check the style guide again. Will change!
</blockquote></details>

Done.

Just to add to this: the use of `"$@"` inside of `[[ ]]` is discouraged by shellcheck, because it performans implicit array concatenation, see https://github.com/koalaman/shellcheck/wiki/SC2199

What I failed to correct it to previously was `"$*"`, which to my understanding would be inline with the style guide – as opposed to the unquoted `$*`.

I now eventually changed it to `(( $# == 0 ))`: I think that expresses the intent best, i.e. that zero arguments had been given. (Quotes are absent due to integer / arithmetic comparison.)
___
*[files/change-hostname, line 46 at r4](https://reviewable.io/reviews/mtlynch/ansible-role-tinypilot/94#-MTa0rwq1lQrQ_78iJUU:-MTkbZ929RL4ikbdU2uS:by3uhyp) ([raw file](https://github.com/mtlynch/ansible-role-tinypilot/blob/8cca6293647ff7b14263a5aafff3f3e57b4af114/files/change-hostname#L46)):*
<details><summary><i>Previously, jotaen (Jan Heuermann) wrote…</i></summary><blockquote>

The style guide says to NOT curly-brace default variables like `"$1"` unless you have a special case like `"${1}2"`

https://google.github.io/styleguide/shellguide.html#variable-expansion

You are right about the `readonly` and the all-caps of course.
</blockquote></details>

Done for the `readonly` and uppercase.

`"${1}"` is [discouraged in the style guide](https://google.github.io/styleguide/shellguide.html#variable-expansion), so I left it as `"$1"`. Or would you prefer the former regardless?
___
*[files/change-hostname, line 47 at r4](https://reviewable.io/reviews/mtlynch/ansible-role-tinypilot/94#-MTa1gGaEsFJu7uLoXBq:-MTkbohPAtnLugZUwx5w:b8bo6yw) ([raw file](https://github.com/mtlynch/ansible-role-tinypilot/blob/8cca6293647ff7b14263a5aafff3f3e57b4af114/files/change-hostname#L47)):*
<details><summary><i>Previously, jotaen (Jan Heuermann) wrote…</i></summary><blockquote>

Starting with bash 3.2 there is a functional difference between `[[ "$foo" =~ "$pattern" ]]` (string matching) and `[[ "$foo" =~ $pattern ]]` (regex matching). I think the style guide doesn’t cover that case.

https://stackoverflow.com/questions/218156/bash-regex-with-quotes

Sorry about the `$is_in_marker_section` – up until just now I thought there exist boolean literals in bash… :facepalm: My bad! (ditto below for the assignment)
</blockquote></details>

Done for the variables (quoting, case, `readonly`).

As mentioned, `=~ "$PATTERN"` performs a string comparison in bash 3.2+. The only option I see to make this syntax happen is to turn on bash 3.1 compatibility mode via `shopt -s compat31`, but I’m not sure that’s a good idea? See also https://google.github.io/styleguide/shellguide.html#tests
___
*[files/change-hostname, line 60 at r4](https://reviewable.io/reviews/mtlynch/ansible-role-tinypilot/94#-MTa2Hlm3IbPjompuqfK:-MTkdCaT5Z5bqQu6bn8K:bc56efr) ([raw file](https://github.com/mtlynch/ansible-role-tinypilot/blob/8cca6293647ff7b14263a5aafff3f3e57b4af114/files/change-hostname#L60)):*
<details><summary><i>Previously, jotaen (Jan Heuermann) wrote…</i></summary><blockquote>

:+1: (see comment about boolean literals)
</blockquote></details>

Done. I also noticed that I hadn’t consistently used single quotes for some other “non-substitution” strings, so that’s also changed elsewhere now.
___
*[files/change-hostname, line 79 at r4](https://reviewable.io/reviews/mtlynch/ansible-role-tinypilot/94#-MTa1FZF1SYu7i9FB4m_:-MTkdDWW2gJsJn7qkQcH:b-896fix) ([raw file](https://github.com/mtlynch/ansible-role-tinypilot/blob/8cca6293647ff7b14263a5aafff3f3e57b4af114/files/change-hostname#L79)):*
<details><summary><i>Previously, jotaen (Jan Heuermann) wrote…</i></summary><blockquote>

:thumbsup: 
</blockquote></details>

Done.

mtlynch (changes requested):
> *[Reviewable](https://reviewable.io/reviews/mtlynch/ansible-role-tinypilot/94)* status: 0 of 2 files reviewed, 1 unresolved discussion (waiting on @jotaen)

___
*[files/change-hostname, line 46 at r4](https://reviewable.io/reviews/mtlynch/ansible-role-tinypilot/94#-MTa0rwq1lQrQ_78iJUU:-MTkmVEb566KcBs_Z0ae:b-l5tmvy) ([raw file](https://github.com/mtlynch/ansible-role-tinypilot/blob/8cca6293647ff7b14263a5aafff3f3e57b4af114/files/change-hostname#L46)):*
<details><summary><i>Previously, jotaen (Jan Heuermann) wrote…</i></summary><blockquote>

Done for the `readonly` and uppercase.

`"${1}"` is [discouraged in the style guide](https://google.github.io/styleguide/shellguide.html#variable-expansion), so I left it as `"$1"`. Or would you prefer the former regardless?
</blockquote></details>

Oh, I missed that "`${1}"` was discouraged. `"$1"` looks like the right way to go.
___
*[files/change-hostname, line 47 at r4](https://reviewable.io/reviews/mtlynch/ansible-role-tinypilot/94#-MTa1gGaEsFJu7uLoXBq:-MTkmosg4DWvTfbGvFm7:b-6mu9tv) ([raw file](https://github.com/mtlynch/ansible-role-tinypilot/blob/8cca6293647ff7b14263a5aafff3f3e57b4af114/files/change-hostname#L47)):*
<details><summary><i>Previously, jotaen (Jan Heuermann) wrote…</i></summary><blockquote>

Done for the variables (quoting, case, `readonly`).

As mentioned, `=~ "$PATTERN"` performs a string comparison in bash 3.2+. The only option I see to make this syntax happen is to turn on bash 3.1 compatibility mode via `shopt -s compat31`, but I’m not sure that’s a good idea? See also https://google.github.io/styleguide/shellguide.html#tests
</blockquote></details>

Ah, I see. Okay, let's stick with unqouted vars for regex matching.
___
*[files/change-hostname, line 13 at r5](https://reviewable.io/reviews/mtlynch/ansible-role-tinypilot/94#-MTkoms0A7bkttduDD_6:-MTkoms0A7bkttduDD_7:bb0vqle) ([raw file](https://github.com/mtlynch/ansible-role-tinypilot/blob/06bd69c989d9ab6640d824c5227c0fa149dd46ac/files/change-hostname#L13)):*
> ```
> readonly VALID_HOSTNAME_PATTERN='^[-0-9A-Za-z]+$'
> readonly MARKER_START='# AUTOGENERATED BY TINYPILOT <<<<<'
> readonly MARKER_END='# >>>>> AUTOGENERATED BY TINYPILOT'
> ```

I just realized testing this that the strings don't make it totally obvious which marker is the start and which is the end. How about this?

```bash
readonly MARKER_START='# --- BEGIN AUTOGENERATED BY TINYPILOT ---'
readonly MARKER_END='# --- END AUTOGENERATED BY TINYPILOT ---'
```

mtlynch (approved):
:lgtm:  thanks!

> *[Reviewable](https://reviewable.io/reviews/mtlynch/ansible-role-tinypilot/94#-:-MTksS-d7ypzXnz-Enof:btapcqv)* status: 0 of 2 files reviewed, 1 unresolved discussion (waiting on @jotaen)

jotaen (commented):
> *[Reviewable](https://reviewable.io/reviews/mtlynch/ansible-role-tinypilot/94)* status: 0 of 2 files reviewed, 1 unresolved discussion (waiting on @mtlynch)

___
*[files/change-hostname, line 13 at r5](https://reviewable.io/reviews/mtlynch/ansible-role-tinypilot/94#-MTkoms0A7bkttduDD_6:-MTkpU2uDsCOcMzNAl2N:bsrdnja) ([raw file](https://github.com/mtlynch/ansible-role-tinypilot/blob/06bd69c989d9ab6640d824c5227c0fa149dd46ac/files/change-hostname#L13)):*

I just realized testing this that the strings don't make it totally obvious which marker is the start and which is the end. How about this?

```bash
readonly MARKER_START='# --- BEGIN AUTOGENERATED BY TINYPILOT ---'
readonly MARKER_END='# --- END AUTOGENERATED BY TINYPILOT ---'
```
</blockquote></details>

Done.

Good idea, I like that better too! I used the following, because that aligns a bit better visually

```
# --- AUTOGENERATED BY TINYPILOT - START ---
127.0.1.1 tinypilot
# --- AUTOGENERATED BY TINYPILOT - END ---
```

PR #103, opened 2021-02-27
Make usb-gadget recyclable

WarheadsSE:
## Summary

Add a script and include it within systemd unit to allow recycling of the USB gadget. This may not have immediate benefit for users, but will enable faster development to any HID/device configuration in the future.

## Details

1. Add `remove-usb-gadget` script, which follows kernel docs, item 6 & 7.
See https://www.kernel.org/doc/Documentation/usb/gadget_configfs.txt
1. Modify tasks/install_usb_gadget to install this script as well
1. Update the systemd service file to allow up/down state, and call this
new script on stop. Now we can `systemctl start|stop|restart usb-gadget`

@mtlynch this was really to scratch an itch when debugging / developing the HID descriptor changes for this CKlau KVM, but figured it was useful enough. :smile_cat:


WarheadsSE:
This may close https://github.com/mtlynch/ansible-role-tinypilot/issues/19 :thinking:

mtlynch:
This is really cool. Thanks!

I'll review when I get back to work on Monday.

mtlynch:
LGTM, thanks!

PR #108, opened 2021-03-04
Always restart the TinyPilot systemd service.

### LINKED ISSUE #86: Configure TinyPilot service to keep retrying on failure
If TinyPilot doesn't have internet access when booting, Flask fails with `dns.resolver.NoResolverConfiguration: Resolver configuration could not be read or specified no nameservers.`.

I thought setting `Restart=always` would do the trick:

https://github.com/mtlynch/ansible-role-tinypilot/blob/a4e195ab6ee349a2ab2d66ded1255d6d1d0c1784/templates/tinypilot.systemd.j2#L17

But it looks like it's more complicated:

https://serverfault.com/a/845473

jdeanwallace:
This PR disables the rate limiting applied to failed systemd service starts.

According to the [systemd.unit docs](https://www.freedesktop.org/software/systemd/man/systemd.unit.html#StartLimitIntervalSec=interval):
> Units which are started more than *burst* times within an *interval* time interval are not permitted to start any more.
>...
>set it to 0 to disable any kind of rate limiting

So `Restart=always` together with `StartLimitIntervalSec=0` will restart the service no matter what the exit code is.

[Demo video.](https://www.loom.com/share/8a7fe589d06449818805c8451ed09884)

Fixes #86

### DIFF
--- a/templates/tinypilot.systemd.j2
+++ b/templates/tinypilot.systemd.j2
@@ -1,6 +1,7 @@
 [Unit]
 Description=TinyPilot - RPi-based virtual KVM
 After=syslog.target network.target
+StartLimitIntervalSec=0
 
 [Service]
 Type=simple


mtlynch:
LGTM, thanks!

PR #109, opened 2021-03-08
Quietly collect debug logs.

### LINKED ISSUE #106: Make collect-debug-logs -q quieter
The [`collect-debug-logs` script](https://github.com/mtlynch/ansible-role-tinypilot/blob/6b89563d3c932f694e7a6ce98add033f1c2e8e63/files/collect-debug-logs) script dumps different TinyPilot-related debug logs to stdout. In addition, it prints status messages about what logs it's retrieving, which also used to go to stdout (until #100).

The problem was that for scripts retrieving the logs programmatically, the status messages cluttered the actual logs. I tried to fix this in a hacky way in #100, but now the status messages are cluttering the TinyPilot log itself whenever the user retrieves the logs through the web interface.

We should revert #100 and replace it with an implementation that has a bash function that prints the progress messages to stdout, but when `-q` is passed, doesn't print any status messages. Error messages should still go to stderr.

jdeanwallace:
Don't print any status/info messages to stdout when running the `collect-debug-logs` script in `-q` silent mode.

# Design Decisions

1. Added a helper function, `print_info`, that prints to stdout only when `collect-debug-logs` is in silent mode.
1. The helper function wraps `printf` and not `echo` in order to have more flexibility when wanting to print special characters (i.e newlines). The caveats include:
    1. You always need to add a newline to the end of each status/info message.
    1. Watchout when printing `--`, which has a special meaning for `printf` (i.e ignore everything after `--`).

[Demo video](https://www.loom.com/share/c6b5cd5c83cd468e9b8ee5a1c45f9d33).

Reverts #100 and Fixes #106

### DIFF
--- a/files/collect-debug-logs
+++ b/files/collect-debug-logs
@@ -13,6 +13,21 @@ Collect debug logs.
 EOF
 }
 
+#######################################
+# Prints a string if not in silent mode.
+# Globals:
+#   FLAG_SILENT_MODE
+# Arguments:
+#   printf arguments
+# Outputs:
+#   Writes a formatted string to stdout.
+#######################################
+print_info() {
+  if [ "$FLAG_SILENT_MODE" != "true" ]; then
+    printf "$@"
+  fi
+}
+
 FLAG_SILENT_MODE=false
 
 # Parse command-line arguments.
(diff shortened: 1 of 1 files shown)


mtlynch:
LGTM, thanks!

PR #120, opened 2021-03-24
Create convenience script for updating uStreamer settings

### LINKED ISSUE #110: Create convenience script for updating uStreamer settings
It's currently fairly tedious to change uStreamer settings and apply them. You can edit the `ustreamer.service` file directly, but any changes will get overwritten by the next update. The current way to persist settings is:

1. Set values in `ustreamer.service`
1. `systemctl daemon-reload && sudo service ustreamer restart`
1. Set the same variables in `/home/tinypilot/settings.yml` so that they'll persist through the next update

We should add a convenience script that reads settings from `/home/tinypilot/settings.yml` (similar to [how `quick-install` does it](https://github.com/mtlynch/tinypilot/blob/5427fccb0fe6fe66460f5b243b485c22c9d29aed/quick-install#L57L67)) and then run a playbook with only `ansible-role-ustreamer`.

The playbook should:

* Skip fact gathering to save time (`gather_facts: no`)
* Run only tags related to ustreamer's systemd settings
  * We'll have to edit ansible-role-ustreamer to add those tags

The script should support a quiet mode so that we can call it programmatically from the frontend later.

---

Update: 2021-03-24

The desired flow for this script should be:

1. User updates `/home/tinypilot/settings.yml` with their desired uStreamer settings
1. User runs `/opt/tinypilot-privileged/update-video-settings`
1. `update-video-settings` (via Ansible) updates uStreamer's systemd file, reloads it, and restarts the uStreamer service

The script should only support changes to uStreamer role variables that affect uStreamer's runtime flags [in the systemd file](https://github.com/mtlynch/ansible-role-ustreamer/blob/57c424b9ba521393eb97e281f68bd46637c1f284/templates/ustreamer.systemd.j2). We shouldn't try to support every possible uStreamer role variable.

Also, we want to execute the uStreamer role, not the TinyPilot role. We really just want to execute the systemd play and then the two handlers come after it to reload and restart ustreamer. I'm hoping we can get this script to complete in <3 seconds.

I'm thinking the call to Ansible can be something like this:

```bash
readonly TINYPILOT_ROLE_NAME="mtlynch.ustreamer"
echo "- hosts: localhost
  connection: local
  become: true
  become_method: sudo
  gather_facts: no
  roles:
    - role: ${TINYPILOT_ROLE_NAME}" > update-settings.yml

ansible-playbook \
  update-settings.yml \
  --inventory localhost, \
  --extra-vars "@/home/tinypilot/settings.yml" \
  --extra-vars "@/home/ustreamer/config.yml" \
  --tags "config"
```

## Recovering settings from install

We have to work around [this hackery](https://github.com/mtlynch/tinypilot/blob/5ad43ceea18ab26ef4ce4889b3414426677e3cd3/quick-install#L29L47) because it means we lose a clean record of how uStreamer was provisioned. I think we can work around that by updating `quick-install` so that instead of setting those to transient environment variables, it appends the appropriate role variables in `/home/tinypilot/settings.yml`.

jdeanwallace:
This PR does 2 things:
1. This forces ansible to dump any custom ansible variables (i.e. not equal to their defaults) to a settings file.
2. Adds a script to change ustreamer settings (and eventually any settings) by running plays that change config files.

The reason why I needed to do #1 instead of just running ansible-role-ustreamer directly, is that it has no idea of the [install vars set in the quick-install script](https://github.com/mtlynch/tinypilot/blob/master/quick-install#L29-L47) and [custom vars set in the ansible-role-tinypilot ustreamer task](https://github.com/mtlynch/ansible-role-tinypilot/blob/master/tasks/ustreamer.yml#L5-L11). This results in the ustreamer using non-tinypilot default values.

The only way I can think of solving this is by making ansible-role-tinypilot dump the settings that was used to provision the device, to the settings file (`/home/tinypilot/settings.yml`). That way any other ansible playbook can read the settings and maintain the correct settings.

@mtlynch based on your feedback over email, the caller of the playbook can still decide to pass in `--extra-vars` via the cli (including the settings file dump of the previous provision).

The reason why I still think this might be a good option is because:
1. With this change, ansible-role-tinypilot doesn't read in any (additional) settings file and the playbook caller still has control over what settings it passes the role.
2. We use ansible to provision the device which means all the decisions to be made around which config to apply should live with ansible.
3. That being said, the TinyPilot application can still make changes to the settings file and it will decide which settings to pass to ansible again.

# Outstanding changes
1. Add quiet mode to convenience script
2. Finalize [tags in ansible-role-ustreamer](https://github.com/mtlynch/ansible-role-ustreamer/pull/37)
3. Fix a bug where there is no video stream in the web UI

# Issues
1. Debugging ansible tasks all day is very time consuming.
1. I often find myself in a "nothing is working & I'm not sure what changed" state when fiddling with ansible settings. I then reinstall from `master` and try again.

Fixes #110

---
This change is [Reviewable](https://reviewable.io/reviews/mtlynch/ansible-role-tinypilot/120)


mtlynch:
Sorry, I think I should have done a better job specifying it.

The thing we need is actually much simpler than what this PR does. We shouldn't need to run the tinypilot Ansible role at all. I think the script works if we just execute the one play we need and its two subsequent handlers.

I updated the [bug description](https://github.com/mtlynch/ansible-role-tinypilot/issues/110#issue-826310419) to provide more details.

The good news is that I think the version we need is going to be *way* less tedious to test because it should run in a few seconds.

jdeanwallace:
Okay sorry about the confusion. I went back to basics and just did what the task said 😅 

>Fix a bug where there is no video stream in the web UI

I still had this bug after reverting (some of) my code. The issue was that the ansible uStreamer role did not have default values that work out the box. Specifically the [`ustreamer_interface` & `ustreamer_port`](https://github.com/mtlynch/ansible-role-ustreamer/blob/master/defaults/main.yml#L8-L12). We only set this in the [ansible TinyPilot role](https://github.com/mtlynch/ansible-role-tinypilot/blob/master/tasks/ustreamer.yml#L5-L11), so running the uStreamer role directly caused the video to break.

[Demo video](https://www.loom.com/share/070e1df0315c46a7bb3cf7954444c56b).

mtlynch (approved):
:lgtm:

Thanks!

> *[Reviewable](https://reviewable.io/reviews/mtlynch/ansible-role-tinypilot/120#-:-MWfOqPP2SnHqYih2kdl:b9benhj)* status: 0 of 3 files reviewed, 4 unresolved discussions (waiting on @jdeanwallace)

___
*[files/update-video-settings, line 36 at r1](https://reviewable.io/reviews/mtlynch/ansible-role-tinypilot/120#-MWfNylJ6wvU0VxTAYlE:-MWfNylJ6wvU0VxTAYlF:bru97gr) ([raw file](https://github.com/mtlynch/ansible-role-tinypilot/blob/1689655f041bde3b6cf3a9c57159e2a3da190e10/files/update-video-settings#L36)):*
> ```
>             print_help >&2
>             exit 1
>         # Add more options in the future.
> ```

We can remove this line. I think we started this convention by mistake and it keeps getting copied.
___
*[files/update-video-settings, line 41 at r1](https://reviewable.io/reviews/mtlynch/ansible-role-tinypilot/120#-MWfOOQ80cU4rJWkioW8:-MWfOOQ80cU4rJWkioW9:bgl5kii) ([raw file](https://github.com/mtlynch/ansible-role-tinypilot/blob/1689655f041bde3b6cf3a9c57159e2a3da190e10/files/update-video-settings#L41)):*
> ```
> 
> 
> if [ ${FLAG_QUIET_MODE} = true ]; then
> ```

We should do double square brackets:

https://google.github.io/styleguide/shellguide.html#s6.3-tests

(ditto throughout)
___
*[files/update-video-settings, line 41 at r1](https://reviewable.io/reviews/mtlynch/ansible-role-tinypilot/120#-MWfOYjl0A11VP9lsslc:-MWfOYjl0A11VP9lssld:bmd97m3) ([raw file](https://github.com/mtlynch/ansible-role-tinypilot/blob/1689655f041bde3b6cf3a9c57159e2a3da190e10/files/update-video-settings#L41)):*
> ```
> 
> 
> if [ ${FLAG_QUIET_MODE} = true ]; then
> ```

We should quote both sides of the comparison:

https://google.github.io/styleguide/shellguide.html#quoting
___
*[files/update-video-settings, line 43 at r1](https://reviewable.io/reviews/mtlynch/ansible-role-tinypilot/120#-MWfO1yU4aoDnvVWbblt:-MWfO1yU4aoDnvVWbblu:b-2b5pvw) ([raw file](https://github.com/mtlynch/ansible-role-tinypilot/blob/1689655f041bde3b6cf3a9c57159e2a3da190e10/files/update-video-settings#L43)):*
> ```
> if [ ${FLAG_QUIET_MODE} = true ]; then
>   # Ignore stdout.
>   exec 1> /dev/null
> ```

(no action required) Oh, neat! I didn't know you could do that.

PR #122, opened 2021-03-31
Move tinypilot_user var from defaults to vars

mtlynch:
This gives the variable higher precedence and indicates to users that they shouldn't change the value.


jdeanwallace (approved):
Looks good. LGTM!

PR #125, opened 2021-04-02
Configure mass storage and add scripts for mounting

jotaen:
I was familiarising myself with configfs and mounting a mass storage today. (So far it’s the basic setup and a script for mounting a flash drive, but it’s all still WIP.)

One question I was running into: we want to support both CD-ROM and flash-drive mode, right? (See [the `cdrom` option](https://www.kernel.org/doc/html/latest/usb/mass-storage.html#module-parameters).) Just double-checking, because CD-ROMs seem to be a bit more “unruly” code-wise, so e.g. swapping out the backing file appears to require rebinding the gadget to the UDC, see [here](https://github.com/tiny-pilot/tinypilot-pro/pull/1/files#diff-48f39e05f5ab360b86cceb5f82c2a2a13f1e2c6cad14601b4b362698bad1af0bR28). Not super bad, but in case we wouldn’t want it it’s a bit simpler.

### DIFF
--- a/files/init-usb-gadget
+++ b/files/init-usb-gadget
@@ -122,6 +122,17 @@ echo -ne \\xC0           # END_COLLECTION
 } >> "$D"
 cp "$D" "${MOUSE_FUNCTIONS_DIR}/report_desc"
 
+# Mass Storage
+# See here for more info:
+# - https://www.kernel.org/doc/html/latest/usb/mass-storage.html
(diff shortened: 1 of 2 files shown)


jotaen:
Ah, and another question (sorry if we had discussed that already): is mass storage pro-exclusive or community as well?

mtlynch:
Yeah, this is for Pro. Can we move the PR there?

Also, could we include an example command in the header comment for `mass-storage-mount`? I'm not clear from the description what kind of argument we pass to the script.

PR #141, opened 2021-06-25
Move uStreamer variables out of tasks

### LINKED ISSUE #140: Move uStreamer variables to defaults/main.yml
We're currently hardcoding the uStreamer role variables in tasks/ustreamer.yml:

https://github.com/tiny-pilot/ansible-role-tinypilot/blob/77700969294ae0d166f66e55f094ede20a2e72d0/tasks/ustreamer.yml#L6L9

It would be more flexible to define those variables in `defaults/main.yml` because that would let role clients override them and we'd avoid implicitly duplicating the values in tasks/nginx.yml:

https://github.com/tiny-pilot/ansible-role-tinypilot/blob/77700969294ae0d166f66e55f094ede20a2e72d0/tasks/nginx.yml#L14L21

jdeanwallace:
This was trickier than expected. I wasn't able to move the ustreamer vars to `defaults/main.yml` and instead I moved them to `vars/main.yml`. The reason being is [ansible's variable precedence](https://docs.ansible.com/ansible/latest/user_guide/playbooks_variables.html#understanding-variable-precedence).

A role's default variables has the lowest priority. So defining variables in `ansible-role-tinypilot/defaults/main.yml` has the same priority level as variables in `ansible-role-ustreamer/defaults/main.yml`. And the [default variables of the last role to be imported takes precedence](https://docs.ansible.com/ansible/latest/user_guide/playbooks_variables.html#id13).

So defining ustreamer variables in `ansible-role-tinypilot/defaults/main.yml` just didn't take effect. But because `ansible-role-tinypilot/vars/main.yml` has a higher priority, it finally worked. This change still allows us to avoid redefining variables inside tasks.

[Demo video](https://www.loom.com/share/7877d5a1c80a488a8f16f877f2709f4e).

Fixes #140

---
This change is [Reviewable](https://reviewable.io/reviews/tiny-pilot/ansible-role-tinypilot/141)

### DIFF
--- a/tasks/main.yml
+++ b/tasks/main.yml
@@ -1,6 +1,8 @@
 ---
 - name: install uStreamer
   import_tasks: ustreamer.yml
+  tags:
+    - ustreamer
 
 - name: install nginx
   import_tasks: nginx.yml
--- a/tasks/nginx.yml
+++ b/tasks/nginx.yml
@@ -11,7 +11,6 @@
   import_role:
     name: ansible-role-nginx
   vars:
-    ustreamer_port: 8001
     nginx_upstreams:
       - name: tinypilot
         servers:
--- a/tasks/ustreamer.yml
+++ b/tasks/ustreamer.yml
@@ -2,8 +2,3 @@
 - name: import uStreamer role
   import_role:
     name: ansible-role-ustreamer
-  vars:
-    ustreamer_interface: '127.0.0.1'
-    ustreamer_port: 8001
-    ustreamer_repo: https://github.com/tiny-pilot/ustreamer.git
-    ustreamer_repo_version: "v3.26"
--- a/vars/main.yml
+++ b/vars/main.yml
@@ -3,3 +3,8 @@
 # relies on the tinypilot user being named "tinypilot" so changing this value
 # will break updates.
 tinypilot_user: tinypilot
+
+ustreamer_interface: '127.0.0.1'
+ustreamer_port: 8001
+ustreamer_repo: https://github.com/tiny-pilot/ustreamer.git
+ustreamer_repo_version: 'v3.26'
--- a/vars/main.yml
+++ b/vars/main.yml
@@ -4,6 +4,9 @@
 # will break updates.
 tinypilot_user: tinypilot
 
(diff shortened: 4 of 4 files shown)


mtlynch (approved):
:lgtm: 

Thanks!

> *[Reviewable](https://reviewable.io/reviews/tiny-pilot/ansible-role-tinypilot/141#-:-MdHhrqu2DJcojAItM_C:b-41s8zb)* status: 0 of 4 files reviewed, 1 unresolved discussion (waiting on @jdeanwallace)

___
*[vars/main.yml, line 7 at r1](https://reviewable.io/reviews/tiny-pilot/ansible-role-tinypilot/141#-MdHhiSaDXZQYHh-n3lj:-MdHhiSaDXZQYHh-n3lk:b-fm9atn) ([raw file](https://github.com/tiny-pilot/ansible-role-tinypilot/blob/0c0c53a67a5e359566459c53197c78b05bc09a93/vars/main.yml#L7)):*
> ```YAML
> tinypilot_user: tinypilot
> 
> ustreamer_interface: '127.0.0.1'
> ```

Can we add a comment above this group just explaining why it's in `vars/` rather than `defaults/`? I can imagine myself forgetting the reasoning and trying to move it to `defaults/` at some point in the future.

jdeanwallace (commented):
> *[Reviewable](https://reviewable.io/reviews/tiny-pilot/ansible-role-tinypilot/141)* status: 0 of 4 files reviewed, 1 unresolved discussion (waiting on @mtlynch)

___
*[vars/main.yml, line 7 at r1](https://reviewable.io/reviews/tiny-pilot/ansible-role-tinypilot/141#-MdHhiSaDXZQYHh-n3lj:-MdIuvBR8KZnBC1FlpT3:b-896fix) ([raw file](https://github.com/tiny-pilot/ansible-role-tinypilot/blob/0c0c53a67a5e359566459c53197c78b05bc09a93/vars/main.yml#L7)):*

Can we add a comment above this group just explaining why it's in `vars/` rather than `defaults/`? I can imagine myself forgetting the reasoning and trying to move it to `defaults/` at some point in the future.
</blockquote></details>

Done.

Issue #144, opened 2021-07-16
Prevent browsers from re-downloading the same static files repeatedly


mtlynch:
Splitting off from https://github.com/tiny-pilot/tinypilot/issues/538#issuecomment-881484720

TinyPilot's nginx configuration [disables caching](https://github.com/tiny-pilot/ansible-role-tinypilot/blob/c565b6020c419db31d3fb94c86f5cf019c56908c/tasks/nginx.yml#L62) for static resources. I originally did this because it put the app into an inconsistent state after the user applied updates (e.g., they'd reload after an update and have different versions of various static files). But we realized that it has the unintended consequence of forcing the browser to redownload the same file dozens of times for a single page load on Chrome (and possibly other browsers).

We should revise the cache policy so that updates don't throw the frontend into an inconsistent state but we're also not causing bizarre browser behavior of redownloading the same file.

jotaen:
Unfortunately, behaviour is wildly inconsistent across browsers, so this issue requires the following fixes. All in all, I’ve tested Firefox and Chrome, on Windows 10, Linux and MacOS. (I’ve also tested MS Edge on Win10, but that’s Chromium-based nowadays, so it behaves the same as Chrome anyway.)

## Part 1: Repeated CSS (Chrome)
This is the biggest issue and it’s happening on Chrome (both Win and MacOS).

Chrome issues one request per `@import` statement (in the CSS) encountered. In our web components, there is no way around importing the CSS individually, since the shadow DOM’s CSS doesn’t inherit from the global scope. That’s why we can’t globally import the CSS upfront. The easiest solution is to turn on caching, as proposed in https://github.com/tiny-pilot/ansible-role-tinypilot/pull/145

I’ve also looked into other ways that don’t rely on heuristics, but that would result in more complexity:

- The most common solution is to put a version number either into the file name itself, or into the query string, and then allow indefinite caching per file version. The biggest problem here is how to embedd these version numbers into the HTML. For that we’d actually need bundling first, then we could come back to this approach.
- As an alternative, we could rely on `ETag` or `If-Modified-Since` headers, which nginx would give us for free. That would still make the browser trigger the requests, though, even if there is no data transmitted in the end because the server sends back a `304 Not Modified`.

## Part 2: Repeated JS (Firefox on MacOS)
This is currently a minor issue and I can only see it on Firefox on MacOS.

[...]

PR #145, opened 2021-07-19
Fix asset caching to avoid redundant requests

jotaen:
Fixes part 1 of https://github.com/tiny-pilot/ansible-role-tinypilot/issues/144.

The resulting behaviour should be that Chrome doesn’t do redundant requests for CSS files anymore. As discussed in the linked issue, we can achieve this via caching. We must avoid by all means, however, that the browser serves from the cache after users perform a device udpate, because that can lead to funky behaviour client-side. Therefore, the caching duration is intentionally short.

## Remarks
- Remove unnecessary caching directives:
  - The `Expires` header will be ignored anyways if `max-age` is present, so we don’t need to configure it. ([Source](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Expires))
  - If the [`etag`](http://nginx.org/en/docs/http/ngx_http_core_module.html#etag) and [`if_modified_since`](http://nginx.org/en/docs/http/ngx_http_core_module.html#if_modified_since) directives are not specified, nginx will enable both by default. I don’t see any harm in doing this: if the browser wants to probe the server to check whether there is something new, we can just let that happen.
- Setting the `Last-Modified` to a recent time is actually crucial, [due to heuristic caching strategies](https://www.w3.org/Protocols/rfc2616/rfc2616-sec13.html#sec13.2.2). The longer a resource hasn’t been modified, the more likely it is for the browser to defer cache invalidation. Interestingly, the `max-age` value alone doesn’t seem to be enough to enforce cache invalidation reliably, I could observe this when testing.
- I set `Cache-Control` to [`public`](https://www.w3.org/Protocols/rfc2616/rfc2616-sec14.html#sec14.9.1) because I think it should be okay for assets to be stored in a shared client-side cache. The assets are not user-specific after all.

## Measurement of transferred assets

(`.css`, `.js` and `.woff2` files; for Pro.)

|  | Before | After |
|--|--|--|
| Number of requests | 128 | 21 |
| kB transferred | 503 kB | 238 kB |

## Sample response (after)
```
GET http://tinypilot/css/cursors.css

Accept-Ranges: bytes
Cache-Control: public, max-age=10s
Connection: keep-alive
Content-Length: 671
Content-Type: text/css
Date: Mon, 19 Jul 2021 08:12:37 GMT
ETag: "60e69367-29f"
Last-Modified: Monday, 19-Jul-2021 08:12:37 GMT
Server: nginx/1.14.2
```

## Sample response (before)
```
GET http://tinypilot/css/cursors.css

Accept-Ranges: bytes
Cache-Control: no-store, no-cache, must-revalidate, proxy-revalidate, max-age=0
Connection: keep-alive
Content-Length: 671
Content-Type: text/css
Date: Mon, 19 Jul 2021 08:42:41 GMT
Last-Modified: Monday, 19-Jul-2021 08:42:41 GMT
Server: nginx/1.14.2
```

### DIFF
--- a/tasks/nginx.yml
+++ b/tasks/nginx.yml
@@ -54,7 +54,7 @@
           location / {
             proxy_pass http://tinypilot;
           }
-          location ~* ^/.+\.(html|js|js.map|css)$ {
+          location ~* ^/.+\.(html|js|js.map|css|woff|woff2)$ {
             root "{{ tinypilot_dir }}/app/static";
 
             # Disable caching
--- a/tasks/nginx.yml
+++ b/tasks/nginx.yml
@@ -57,12 +57,14 @@
           location ~* ^/.+\.(html|js|js.map|css|woff|woff2)$ {
             root "{{ tinypilot_dir }}/app/static";
 
-            # Disable caching
+            # We cache assets to prevent the browser from making redundant
+            # requests to the same files while loading the page. (Observed on
+            # Chrome 91.) We don’t want caching otherwise, though, in order to
+            # avoid stale files after users update their device. Note, that in
+            # addition to `max-age`, the browser’s caching behaviour is relative
+            # to the `Last-Modified` header, so we make that seem recent.
             add_header Last-Modified $date_gmt;
-            add_header Cache-Control 'no-store, no-cache, must-revalidate, proxy-revalidate, max-age=0';
-            if_modified_since off;
-            expires off;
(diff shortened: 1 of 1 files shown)


mtlynch (approved):
Oof, wish this was simpler. Thanks for the deep investigation. This looks like the best solution, all things considered.

PR #146, opened 2021-07-19
Serve font files via nginx

jotaen:
While working on https://github.com/tiny-pilot/ansible-role-tinypilot/issues/144, I noticed that [our font files](https://github.com/tiny-pilot/tinypilot/tree/master/app/static/third-party/fonts) are not served via nginx. We actually had just forgotten about this back then, [when we started serving font files locally](https://github.com/tiny-pilot/tinypilot/pull/575).

### DIFF
--- a/tasks/nginx.yml
+++ b/tasks/nginx.yml
@@ -54,7 +54,7 @@
           location / {
             proxy_pass http://tinypilot;
           }
-          location ~* ^/.+\.(html|js|js.map|css)$ {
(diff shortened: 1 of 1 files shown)


mtlynch (approved):
LGTM, thanks!

Nice catch!

jotaen:
@mtlynch unrelated: I noticed we have auto-deletion of branches in the other repos, but not here. You find it in the “Settings” of this repo, the checkmark is called “Automatically delete head branches”.

mtlynch:
@jotaen - Good tip, updated!

PR #155, opened 2021-11-01
Patch keyboard HID module.

### LINKED ISSUE #151: Apple doesn't recognize TinyPilot keyboard in pre-boot
Apple devices don't implement the HID spec correctly, so it fails to recognize the TinyPilot keyboard in pre-boot. The upstream Linux kernel [has a workaround](https://forum.tinypilotkvm.com/-132/keyboard-and-mouse-not-working-with-macbook-pro-usb-c), but it hasn't made its way into Raspbian yet.

A reader on my blog [shared instructions](https://mtlynch.io/key-mime-pi/#comment-35) for compiling custom builds of the kernel modules and confirmed that they work on his Pi Zero + Apple devices:

### Instructions

```bash
git clone https://github.com/raspberrypi/linux/
git checkout 4e5d621498df # ("overlays: Reduce Pi 4 vc4 CMA size to 320MB")
get the patch from https://0bin.net/paste/m6fB-sxx#ZF8Jp5ArPkDJS-W3ax5BROkPtkKi+hUH9F/mhgSL+Z1
git apply 0001-usb-gadget-f_hid-optional-SETUP-SET_REPORT-mode.patch # apply patch

CONFIG_LOCALVERSION="-v7l-MY_CUSTOM_KERNEL" # set some environment variable
KERNEL=kernel

make ARCH=arm CROSS_COMPILE=arm-linux-gnueabihf- bcmrpi_defconfig # default config pi zero
make ARCH=arm CROSS_COMPILE=arm-linux-gnueabihf- modules_prepare # prepare build environment for modules
make ARCH=arm CROSS_COMPILE=arm-linux-gnueabihf- M=drivers/usb/gadget/function/
make ARCH=arm CROSS_COMPILE=arm-linux-gnueabihf- M=drivers/usb/gadget/legacy/

mkdir -p apple-keyboard-fix
cp ./drivers/usb/gadget/legacy/g_hid.ko apple-keyboard-fix/ # grab the two modules
cp ./drivers/usb/gadget/function/usb_f_hid.ko apple-keyboard-fix/
```

>On the actual device/pi zero:

```bash
sudo cp /home/pi/apple-keyboard-fix/g_hid.ko /lib/modules/5.10.52+/kernel/drivers/usb/gadget/legacy/g_hid.ko
sudo cp /home/pi/apple-keyboard-fix/usb_f_hid.ko /lib/modules/5.10.52+/kernel/drivers/usb/gadget/function/usb_f_hid.ko
```

>I've been using the instructions from https://www.isticktoit.net/?p=1383 where you ned to add the following to the script /usr/bin/isticktoit_usb, after subclass configuration:

```bash
echo 1 > functions/hid.usb0/no_out_endpoint
```

>I've uploaded both compiled files to my google drive, if you want to skip the instructions above:
g_hid.ko (for kernel 5.10.52+): https://drive.google.com/file/d/17yx2yxFjIjcXplfqRHKXlQAKavysMoeA/view?usp=sharing
usb_f_hid.ko (for kernel 5.10.52+): https://drive.google.com/file/d/10xkzEgTdCPS0M4D1TMv80tlvLI5sY2Mz/view?usp=sharing

### Our fix

We should compile our own build for the Pi 4 and host it on a public TinyPilot URL.

This Ansible role should include logic that if the system is a Raspberry Pi with a kernel version without the workaround, we should download our custom kernel module builds and replace the built-in modules on the device.

I'm not sure if we need to replace both modules or only one.

jdeanwallace:
The patched HID module allows a Mac to recognize the TinyPilot keyboard in pre-boot (i.e. login screen).

The [patch](https://github.com/raspberrypi/linux/commit/d7428bc26fc767942c38d74b80299bcd4f01e7cb) is already included kernel versions [`>= 5.15`](https://github.com/raspberrypi/linux/commits/rpi-5.15.y/drivers/usb/gadget/function). However, the latest LTS kernel is version `5.10`. This PR installs the patched HID module on kernel versions `>=5.10` and `<5.15`.

This PR does the following:
1. Downloads and installs the patched `usb_f_hid.ko` module ([here](https://github.com/tiny-pilot/ansible-role-tinypilot/blob/b7e7f89983c27bea88452d8748b96fbb2a0db8f7/tasks/install_usb_gadget.yml#L68-L73)).
2. Updates the USB gadget initializer by adding an additional `no_out_endpoint` HID function ([here](https://github.com/tiny-pilot/ansible-role-tinypilot/blob/b7e7f89983c27bea88452d8748b96fbb2a0db8f7/tasks/install_usb_gadget.yml#L75-L84)). This HID function is only supported after the patch. We cannot add this function without the patched module.

Fixes #151

---
This change is [Reviewable](https://reviewable.io/reviews/tiny-pilot/ansible-role-tinypilot/155)

### DIFF
--- a/files/init-usb-gadget
+++ b/files/init-usb-gadget
@@ -72,6 +72,10 @@ D=$(mktemp)
   echo -ne \\xC0            # End Collection
 } >> "$D"
 cp "$D" "${KEYBOARD_FUNCTIONS_DIR}/report_desc"
+# Enable pre-boot events (if the gadget driver supports it).
+if [[ -f "${KEYBOARD_FUNCTIONS_DIR}/no_out_endpoint" ]]; then
+  echo 1 > "${KEYBOARD_FUNCTIONS_DIR}/no_out_endpoint"
+fi
 
 # Mouse
 MOUSE_FUNCTIONS_DIR="functions/hid.mouse"
--- a/handlers/main.yml
+++ b/handlers/main.yml
@@ -15,3 +15,8 @@
     daemon_reload: yes
     # Never run automatically.
     enabled: no
+
+- name: start usb-gadget service
+  service:
+    name: usb-gadget
+    state: started
--- a/tasks/install_usb_gadget.yml
+++ b/tasks/install_usb_gadget.yml
@@ -65,6 +65,24 @@
     mode: '0644'
   register: usb_gadget_template
 
+- name: patch HID module
+  get_url:
+    url: https://storage.googleapis.com/jason-tinypilotkvm/usb_f_hid.ko
+    dest: /lib/modules/{{ ansible_kernel }}/kernel/drivers/usb/gadget/function/usb_f_hid.ko
+    mode: '0664'
+  when: ansible_kernel is version('5.15', '<') and ansible_kernel is version('5.10', '>=')
+
+- name: patch usb-gadget initializer
+  ansible.builtin.lineinfile:
+    path: "{{ tinypilot_privileged_dir }}/init-usb-gadget"
+    line: 'echo 1 > "${KEYBOARD_FUNCTIONS_DIR}/no_out_endpoint"'
+    # Either replace the line this regex is found on.
+    regex: ^{{ 'echo 1 > "${KEYBOARD_FUNCTIONS_DIR}/no_out_endpoint"' | regex_escape() }}
+    # Or insert the line after this regex is found.
+    insertafter: ^{{ 'mkdir -p "$KEYBOARD_FUNCTIONS_DIR"' | regex_escape() }}$
+    # Otherwise add the line to the EOF.
+  when: ansible_kernel is version('5.15', '<') and ansible_kernel is version('5.10', '>=')
+
 - name: enable systemd usb-gadget initializer service file
   systemd:
     name: usb-gadget
--- a/tasks/install_usb_gadget.yml
+++ b/tasks/install_usb_gadget.yml
@@ -81,7 +81,7 @@
     # Or insert the line after this regex is found.
     insertafter: ^{{ 'mkdir -p "$KEYBOARD_FUNCTIONS_DIR"' | regex_escape() }}$
     # Otherwise add the line to the EOF.
-  when: ansible_kernel is version('5.15', '<') and ansible_kernel is version('5.10', '>=')
+  when: ansible_kernel is version('5.10', '>=')
 
 - name: enable systemd usb-gadget initializer service file
   systemd:
--- a/tasks/install_usb_gadget.yml
+++ b/tasks/install_usb_gadget.yml
@@ -65,23 +65,16 @@
     mode: '0644'
   register: usb_gadget_template
 
-- name: patch HID module
+# This custom USB kernel module fixes compatibility issues in pre-boot for OS X
+# systems. The patch already exists in kernal versions >= 5.15.
+# Issue: https://github.com/tiny-pilot/ansible-role-tinypilot/issues/151
+- name: patch USB kernel module
   get_url:
-    url: https://storage.googleapis.com/jason-tinypilotkvm/usb_f_hid.ko
+    url: https://github.com/tiny-pilot/hid-backport/raw/master/bin/usb_f_hid.ko
     dest: /lib/modules/{{ ansible_kernel }}/kernel/drivers/usb/gadget/function/usb_f_hid.ko
     mode: '0664'
+    checksum: sha256:45f9c885e8b0e1d2fdaf4aec2179a38ad9a1c76c71f44c997c99a865fdfe72d7
   when: ansible_kernel is version('5.15', '<') and ansible_kernel is version('5.10', '>=')
-
-- name: patch usb-gadget initializer
-  ansible.builtin.lineinfile:
-    path: "{{ tinypilot_privileged_dir }}/init-usb-gadget"
-    line: 'echo 1 > "${KEYBOARD_FUNCTIONS_DIR}/no_out_endpoint"'
-    # Either replace the line this regex is found on.
-    regex: ^{{ 'echo 1 > "${KEYBOARD_FUNCTIONS_DIR}/no_out_endpoint"' | regex_escape() }}
-    # Or insert the line after this regex is found.
-    insertafter: ^{{ 'mkdir -p "$KEYBOARD_FUNCTIONS_DIR"' | regex_escape() }}$
-    # Otherwise add the line to the EOF.
-  when: ansible_kernel is version('5.10', '>=')
 
 - name: enable systemd usb-gadget initializer service file
   systemd:
--- a/tasks/install_usb_gadget.yml
+++ b/tasks/install_usb_gadget.yml
@@ -65,19 +65,48 @@
     mode: '0644'
   register: usb_gadget_template
 
-# This custom USB kernel module fixes compatibility issues in pre-boot for OS X
-# systems. The patch already exists in kernal versions >= 5.15.
-# Issue: https://github.com/tiny-pilot/ansible-role-tinypilot/issues/151
-- name: patch USB kernel module
-  get_url:
-    url: https://github.com/tiny-pilot/hid-backport/raw/master/bin/usb_f_hid.ko
-    dest: /lib/modules/{{ ansible_kernel }}/kernel/drivers/usb/gadget/function/usb_f_hid.ko
-    mode: '0664'
-    checksum: sha256:45f9c885e8b0e1d2fdaf4aec2179a38ad9a1c76c71f44c997c99a865fdfe72d7
-  when: ansible_kernel is version('5.15', '<') and ansible_kernel is version('5.10', '>=')
-
 - name: enable systemd usb-gadget initializer service file
   systemd:
     name: usb-gadget
     enabled: yes
     daemon_reload: "{{ usb_gadget_template.changed }}"
+
+# This custom USB HID kernel module fixes compatibility issues in pre-boot for
+# OS X systems. The patch already exists in kernal versions >= 5.15.
+# Issue: https://github.com/tiny-pilot/ansible-role-tinypilot/issues/151
+- name: check the HID module file
+  ansible.builtin.stat:
+    path: /lib/modules/{{ ansible_kernel }}/kernel/drivers/usb/gadget/function/usb_f_hid.ko
+    checksum_algorithm: sha256
+  register: hid_module_stat
+
+- name: save whether the HID module should be patched
+  set_fact:
+    patch_hid_module: >-
+      {{ ansible_kernel is version('5.15', '<')
+         and ansible_kernel is version('5.10', '>=')
+         and hid_module_stat.stat.checksum != '45f9c885e8b0e1d2fdaf4aec2179a38ad9a1c76c71f44c997c99a865fdfe72d7' }}
+
+- name: ensure HID module is not in use
+  service:
+    name: usb-gadget
+    state: stopped
+  when: patch_hid_module
+
+- name: unload HID module
+  command: modprobe --remove usb_f_hid
+  when: patch_hid_module
+
+- name: patch HID module
+  get_url:
+    url: https://github.com/tiny-pilot/hid-backport/raw/master/bin/usb_f_hid.ko
+    dest: "{{ hid_module_stat.stat.path }}"
+    mode: '0644'
+    force: yes
+  when: patch_hid_module
+
+- name: load HID module
+  command: modprobe usb_f_hid
+  when: patch_hid_module
+  notify:
+    - start usb-gadget service


jdeanwallace:
@mtlynch can you please upload the [compiled `usb_f_hid.ko` file](https://storage.googleapis.com/jason-tinypilotkvm/usb_f_hid.ko) to a GCS bucket and make is publicly accessible?

mtlynch:
Doesn't 5.13 have the fix? We don't want to mess with the kernel if we don't have to, so I want to be as conservative as possible in patching the module if we don't need to.

I'm not following the logic for using 5.10 as the min version. Doesn't every kernel version before 5.13 need this fix? We're not guaranteed that the user is running the latest Raspbian.

Also, for situations where we're applying a workaround, I'd like to go heavy on code comments and clear naming. We should aim for the reader to understand the context of the workaround without digging into `git blame` and finding associated issues and PRs (though it's good to link to the associated issues).

>@mtlynch can you please upload the compiled usb_f_hid.ko file to a GCS bucket and make is publicly accessible?

Can you share a bash script that compiles it? I want to store the script with the .ko so that we can reproduce it if necessary. The end-user won't see the script. It will just sit in the bucket folder next to the .ko.

jdeanwallace:
> Doesn't 5.13 have the fix?

Not from what I can see. The fix adds support for an additional [`no_out_endpoint` function here](https://github.com/raspberrypi/linux/blob/d7428bc26fc767942c38d74b80299bcd4f01e7cb/drivers/usb/gadget/function/u_hid.h#L23).

* Version [5.13 doesn't have it](https://github.com/raspberrypi/linux/blob/rpi-5.13.y/drivers/usb/gadget/function/u_hid.h).
* Version [5.14 doesn't have it](https://github.com/raspberrypi/linux/blob/rpi-5.14.y/drivers/usb/gadget/function/u_hid.h).
* Version [5.15 does have it](https://github.com/raspberrypi/linux/blob/rpi-5.15.y/drivers/usb/gadget/function/u_hid.h#L23).

---

> I'm not following the logic for using 5.10 as the min version. Doesn't every kernel version before 5.13 need this fix? We're not guaranteed that the user is running the latest Raspbian.

I used 5.10 as the min version because it is latest LTS version that I tested on. I wasn't sure how far back the patch diff would be successfully applied to the file. I didn't think it was that important to be backwards compatible because I thought "hey, if you want the latest features, then be on the latest version".

I just tried to apply the patch diff to earliest kernel version where the patched files exist (i.e. [version 3.19](https://github.com/raspberrypi/linux/blob/rpi-3.19.y/drivers/usb/gadget/function/u_hid.h)), but the patch did no apply successfully. From which kernel version should I start applying this patch? I suggest we only support from the earliest kernel version that TinyPilot Pro was first shipped with.

---

> Also, for situations where we're applying a workaround, I'd like to go heavy on code comments and clear naming. We should aim for the reader to understand the context of the workaround without digging into git blame and finding associated issues and PRs (though it's good to link to the associated issues).

Okay cool, I'll add more explanation to the code that applies the workaround.

---

> Can you share a bash script that compiles it?

Sure, you can run the following script on the device as root:

```bash
#!/bin/bash

# Exit build script on first failure.
set -e

# Echo commands to stdout.
set -x

# Exit on unset variable.
set -u

# Install dependencies.
apt-get install git bc bison flex libssl-dev make libc6-dev libncurses5-dev -y

# Clone repo.
BUILD_DIR="$(mktemp --directory)"
pushd "${BUILD_DIR}"
git clone --depth=1 --branch rpi-5.10.y https://github.com/raspberrypi/linux
pushd linux

# Apply patch.
curl https://github.com/raspberrypi/linux/commit/d7428bc26fc767942c38d74b80299bcd4f01e7cb.patch \
  --output fix.patch
git apply fix.patch

# Generate config.
KERNEL=kernel7l
make ARCH=arm CROSS_COMPILE=arm-linux-gnueabihf- bcm2711_defconfig

# Build module.
make ARCH=arm CROSS_COMPILE=arm-linux-gnueabihf- modules_prepare
make ARCH=arm CROSS_COMPILE=arm-linux-gnueabihf- M=drivers/usb/gadget/function

# Install module.
cp ./drivers/usb/gadget/function/usb_f_hid.ko \
  /lib/modules/$(uname -r)/kernel/drivers/usb/gadget/function/

```

mtlynch:
@jdeanwallace - Thanks for the explanation. Good points! I agree with you on applying it only to >= 5.10 through < 5.15.

The script you shared clones only the the `rpi-5.10.y` branch. We want `rpi-5.15.y`, though, right?

jdeanwallace:
> The script you shared clones only the the rpi-5.10.y branch. We want rpi-5.15.y, though, right?

Bizarrely, that doesn't work 🤔. The module builds fine, but after the install & reboot, the [`init-usb-gadget`](https://github.com/tiny-pilot/ansible-role-tinypilot/blob/b7e7f89983c27bea88452d8748b96fbb2a0db8f7/files/init-usb-gadget) script in unable to create the `/sys/kernel/config/usb_gadget/g1/functions/hid.keyboard` directory. Full output below:

```bash
/opt/tinypilot-privileged/init-usb-gadget
+ set -u
+ modprobe libcomposite
+ cd /sys/kernel/config/usb_gadget/
+ mkdir -p g1
+ cd g1
+ echo 0x1d6b
+ echo 0x0104
+ echo 0x0100
+ echo 0x0200
+ STRINGS_DIR=strings/0x409
+ mkdir -p strings/0x409
+ echo 6b65796d696d6570690
+ echo tinypilot
+ echo 'Multifunction USB Device'
+ KEYBOARD_FUNCTIONS_DIR=functions/hid.keyboard
+ mkdir -p functions/hid.keyboard
mkdir: cannot create directory 'functions/hid.keyboard': No such file or directory
```
I'm actually unable to create any file or directory inside `/sys/kernel/config/usb_gadget/g1/functions/` 🤔 

If I run the script with the `rpi-5.10.y` branch, then it works again after reboot.

mtlynch:
Oh, gotcha. I missed that we were applying a patch too. I'll create a public github repo for this and then check in the compiled .ko.

mtlynch:
Okay, here's the public version:

https://github.com/tiny-pilot/hid-backport/raw/master/bin/usb_f_hid.ko

The repo itself is obviously at:

https://github.com/tiny-pilot/hid-backport

Can you verify that my `usb_f_hid.ko` file works the same as the version you compiled?

jdeanwallace:
Thanks!

> Can you verify that my `usb_f_hid.ko` file works the same as the version you compiled?

Yes, it works!

mtlynch (commented):
> *[Reviewable](https://reviewable.io/reviews/tiny-pilot/ansible-role-tinypilot/155)* status: 0 of 1 LGTMs obtained

___
*[tasks/install_usb_gadget.yml, line 77 at r1](https://reviewable.io/reviews/tiny-pilot/ansible-role-tinypilot/155#-MnfN05c1sKQ260CEjbJ:-MnfN05c1sKQ260CEjbK:b-doad70) ([raw file](https://github.com/tiny-pilot/ansible-role-tinypilot/blob/b7e7f89983c27bea88452d8748b96fbb2a0db8f7/tasks/install_usb_gadget.yml#L77)):*
> ```YAML
> - name: patch usb-gadget initializer
>   ansible.builtin.lineinfile:
>     path: "{{ tinypilot_privileged_dir }}/init-usb-gadget"
> ```

We're overwriting the entire `init-usb-gadget` file in the "copy usb-gadget initializer" so I think it's fine to just update in that file directly instead of copying it and making a really delicate tweak to a particular line. Does anything bad happen if we include the line on < 5.10?

mtlynch (commented):
> *[Reviewable](https://reviewable.io/reviews/tiny-pilot/ansible-role-tinypilot/155)* status: 0 of 1 LGTMs obtained

___
*[tasks/install_usb_gadget.yml, line 69 at r1](https://reviewable.io/reviews/tiny-pilot/ansible-role-tinypilot/155#-MngigT-3gO3wTZ_tTlW:-MngigT01dYqbobr9hWM:b-cbfhom) ([raw file](https://github.com/tiny-pilot/ansible-role-tinypilot/blob/b7e7f89983c27bea88452d8748b96fbb2a0db8f7/tasks/install_usb_gadget.yml#L69)):*
> ```YAML
> 
> - name: patch HID module
>   get_url:
> ```

I'm realizing we'll want something to indicate that we've applied the backport already, otherwise we'll slow down the role by downloading and replacing this file every time.

My thought is that we save a file like `/home/tinypilot/.applied-hid-workaround` on change and skip this play if the file already exists?

We do something sort of similar in uStreamer:

https://github.com/tiny-pilot/ansible-role-ustreamer/blob/f09f0f9b0da4a840d8b588435b0a57cfa27d4319/tasks/main.yml#L100

jdeanwallace (commented):
> *[Reviewable](https://reviewable.io/reviews/tiny-pilot/ansible-role-tinypilot/155)* status: 0 of 1 LGTMs obtained

___
*[tasks/install_usb_gadget.yml, line 69 at r1](https://reviewable.io/reviews/tiny-pilot/ansible-role-tinypilot/155#-MngigT-3gO3wTZ_tTlW:-MnkT6X959V8DgzOfXiy:bwhlc4o) ([raw file](https://github.com/tiny-pilot/ansible-role-tinypilot/blob/b7e7f89983c27bea88452d8748b96fbb2a0db8f7/tasks/install_usb_gadget.yml#L69)):*

I'm realizing we'll want something to indicate that we've applied the backport already, otherwise we'll slow down the role by downloading and replacing this file every time.

My thought is that we save a file like `/home/tinypilot/.applied-hid-workaround` on change and skip this play if the file already exists?

We do something sort of similar in uStreamer:

https://github.com/tiny-pilot/ansible-role-ustreamer/blob/f09f0f9b0da4a840d8b588435b0a57cfa27d4319/tasks/main.yml#L100
</blockquote></details>

Good point. I see that [`get_url` has a `checksum` parameter](https://docs.ansible.com/ansible/latest/collections/ansible/builtin/get_url_module.html#parameter-checksum) that will only download the file if the local file doesn't exist or has a different checksum. I've added that here: https://github.com/tiny-pilot/ansible-role-tinypilot/blob/81465175329e6ee8eb55a730c3f3232f6c3165a2/tasks/install_usb_gadget.yml#L76
___
*[tasks/install_usb_gadget.yml, line 77 at r1](https://reviewable.io/reviews/tiny-pilot/ansible-role-tinypilot/155#-MnfN05c1sKQ260CEjbJ:-MnkVtKb0BVaKf6vSePk:b-kphsy0) ([raw file](https://github.com/tiny-pilot/ansible-role-tinypilot/blob/b7e7f89983c27bea88452d8748b96fbb2a0db8f7/tasks/install_usb_gadget.yml#L77)):*
> Does anything bad happen if we include the line on < 5.10?

If the USB gadget doesn't support the endpoint then it won't allow us to write to the file. However, it seems like you can easily determine if the gadget supports `/no_out_endpoint` by just checking if the file exists. The gadget only supports the endpoints that exist on the filesystem.

I've added code to `init-usb-gadget` that only enables the feature if the file exists:
https://github.com/tiny-pilot/ansible-role-tinypilot/blob/81465175329e6ee8eb55a730c3f3232f6c3165a2/files/init-usb-gadget#L75-L78

mtlynch (changes requested):
> *[Reviewable](https://reviewable.io/reviews/tiny-pilot/ansible-role-tinypilot/155)* status: 0 of 1 LGTMs obtained

___
*[tasks/install_usb_gadget.yml, line 69 at r1](https://reviewable.io/reviews/tiny-pilot/ansible-role-tinypilot/155#-MngigT-3gO3wTZ_tTlW:-MnkdUqb8DsNxtONGuBi:b51wgrl) ([raw file](https://github.com/tiny-pilot/ansible-role-tinypilot/blob/b7e7f89983c27bea88452d8748b96fbb2a0db8f7/tasks/install_usb_gadget.yml#L69)):*

Good point. I see that [`get_url` has a `checksum` parameter](https://docs.ansible.com/ansible/latest/collections/ansible/builtin/get_url_module.html#parameter-checksum) that will only download the file if the local file doesn't exist or has a different checksum. I've added that here: https://github.com/tiny-pilot/ansible-role-tinypilot/blob/81465175329e6ee8eb55a730c3f3232f6c3165a2/tasks/install_usb_gadget.yml#L76
</blockquote></details>

Oh, nice! Very clean.
___
*[tasks/install_usb_gadget.yml, line 77 at r2](https://reviewable.io/reviews/tiny-pilot/ansible-role-tinypilot/155#-MnkdklX9j6m6THnQ13p:-MnkdklX9j6m6THnQ13q:b-7ktye2) ([raw file](https://github.com/tiny-pilot/ansible-role-tinypilot/blob/81465175329e6ee8eb55a730c3f3232f6c3165a2/tasks/install_usb_gadget.yml#L77)):*
> ```YAML
>     mode: '0664'
>     checksum: sha256:45f9c885e8b0e1d2fdaf4aec2179a38ad9a1c76c71f44c997c99a865fdfe72d7
>   when: ansible_kernel is version('5.15', '<') and ansible_kernel is version('5.10', '>=')
> ```

Do we need to add a handler to reload the kernel module if we change it?

jdeanwallace (commented):
> *[Reviewable](https://reviewable.io/reviews/tiny-pilot/ansible-role-tinypilot/155)* status: 0 of 1 LGTMs obtained

___
*[tasks/install_usb_gadget.yml, line 77 at r2](https://reviewable.io/reviews/tiny-pilot/ansible-role-tinypilot/155#-MnkdklX9j6m6THnQ13p:-Mo-5gKt0qNjzMAbtCeO:bdim9il) ([raw file](https://github.com/tiny-pilot/ansible-role-tinypilot/blob/81465175329e6ee8eb55a730c3f3232f6c3165a2/tasks/install_usb_gadget.yml#L77)):*

Do we need to add a handler to reload the kernel module if we change it?
</blockquote></details>

Yes, I was relying on the module to be reloaded after a reboot. However, the reboot happens outside of ansible.

I'm trying to add a handler to reload the module using `modprobe [-r] usb_f_hid`, but it's not always successful because it depends on the state of the `usb-gadget` service (i.e. is the gadget in use).

The safest/simplest option is the `reboot` because it can never fail.

I'll try add extra logic to ansible task to ensure a smooth kernel module reload.

jdeanwallace (commented):
I've managed to safely patch and reload the `usb_f_hid` module.

Just a note on something strange that happens when I attempt to undo the patch. Before the module is patched, it is easy to unload (`modprobe --remove usb_f_hid`) and load (`modprobe usb_f_hid`) the module, as long as the `usb-gadget` is removed. However, after the module is patched I am unable to unload the module. The only way I can reload a new module is to reboot the device. I received the following error message when running `modprobe --remove usb_f_hid` while the patched module is loaded:

```bash
modprobe: ERROR: ../libkmod/libkmod-module.c:793 kmod_module_remove_module() could not remove 'usb_f_hid': Device or resource busy
```

> *[Reviewable](https://reviewable.io/reviews/tiny-pilot/ansible-role-tinypilot/155#-:-Mo3uGvX2j7relmUpBFF:bahj2yl)* status: 0 of 1 LGTMs obtained

___
*[tasks/install_usb_gadget.yml, line 77 at r2](https://reviewable.io/reviews/tiny-pilot/ansible-role-tinypilot/155#-MnkdklX9j6m6THnQ13p:-Mo3s-Z7D9vIGY1phkVw:b-896fix) ([raw file](https://github.com/tiny-pilot/ansible-role-tinypilot/blob/81465175329e6ee8eb55a730c3f3232f6c3165a2/tasks/install_usb_gadget.yml#L77)):*

Yes, I was relying on the module to be reloaded after a reboot. However, the reboot happens outside of ansible.

I'm trying to add a handler to reload the module using `modprobe [-r] usb_f_hid`, but it's not always successful because it depends on the state of the `usb-gadget` service (i.e. is the gadget in use).

The safest/simplest option is the `reboot` because it can never fail.

I'll try add extra logic to ansible task to ensure a smooth kernel module reload.
</blockquote></details>

Done.

PR #162, opened 2021-11-15
Ensure stat checksum is defined.

jdeanwallace:
Ansible's `stat.checksum` attribute only exist under the [following conditions](https://docs.ansible.com/ansible/latest/collections/ansible/builtin/stat_module.html#return-stat/checksum):
> path exists, user can read stats, path supports hashing and supplied checksum algorithm is available

Fixes https://github.com/tiny-pilot/ansible-role-tinypilot-pro/issues/73

---
This change is [Reviewable](https://reviewable.io/reviews/tiny-pilot/ansible-role-tinypilot/162)

### DIFF
--- a/tasks/install_usb_gadget.yml
+++ b/tasks/install_usb_gadget.yml
@@ -92,6 +92,7 @@
     patch_hid_module: >-
       {{ ansible_kernel is version('5.15', '<')
          and ansible_kernel is version('5.10', '>=')
+         and hid_module_stat.stat.checksum is defined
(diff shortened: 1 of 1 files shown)


mtlynch:
Closing this as we're going to focus on Packer, but this fix wouldn't do what we want. It would mean that instead of loudly stopping when it fails to calculate the hash for the kernel module, it would just silently skip the patch. So it would unblock the build, but we'd lose the fix for the HID USB gadget that you backported.

PR #165, opened 2021-12-13
Verify HID module checksum exists before attempting to compare it

### LINKED ISSUE #164: Updating the Linux kernel causes install to fail
A user reported the following failure in the TinyPilot updater:

```text
TASK [ansible-role-tinypilot : save whether the HID module should be patched] ***
fatal: [localhost]: FAILED! => {"msg": "The task includes an option with an undefined variable. The error was: 'dict object' has no attribute 'checksum'\n\nThe error appears to be in '/opt/tinypilot-updater/ansible-role-tinypilot/tasks/install_usb_gadget.yml': line 90, column 3, but may\nbe elsewhere in the file depending on the exact syntax problem.\n\nThe offending line appears to be:\n\n\n- name: save whether the HID module should be patched\n ^ here\n"}
```

The failure is here:

https://github.com/tiny-pilot/ansible-role-tinypilot/blob/e0d6b472796146ae8e7b8a3215ad6cd13d73b561/tasks/install_usb_gadget.yml#L95

The user mentioned that they had just performed system upgrades, and the update worked after rebooting. It looks like if there's a reboot pending a kernel update, the `usb_f_hid.ko` won't be there, which causes the play to fail.

mtlynch:
The HID module patch task can fail if the target system has just upgraded the kernel, as the path to the usb_f_hid.ko won't be correct until the following reboot. To avoid crashing the install with an 'undefined variable' error, we should first verify whether the variables we need are defined.

Fixes #164

---
This change is [Reviewable](https://reviewable.io/reviews/tiny-pilot/ansible-role-tinypilot/165)

### DIFF
--- a/tasks/install_usb_gadget.yml
+++ b/tasks/install_usb_gadget.yml
@@ -92,6 +92,7 @@
     patch_hid_module: >-
       {{ ansible_kernel is version('5.15', '<')
          and ansible_kernel is version('5.10', '>=')
+         and hid_module_stat.stat.checksum is defined
          and hid_module_stat.stat.checksum != '45f9c885e8b0e1d2fdaf4aec2179a38ad9a1c76c71f44c997c99a865fdfe72d7' }}
 
 - name: ensure HID module is not in use


mtlynch:
@jdeanwallace - I think you had proposed this at one point when we found that our image building environment had the same problem with missing `usb_f_hid.ko`, and I didn't want to silently ignore the error. In this case, since it's happening on customer device and blocking updates, it's worth suppressing the failure.

jdeanwallace (approved):
:lgtm: 

> *[Reviewable](https://reviewable.io/reviews/tiny-pilot/ansible-role-tinypilot/165#-:-MqsOYqR6oHYTuIfIxoR:bg7igkv)* status: :shipit: complete! 1 of 1 LGTMs obtained

PR #171, opened 2022-01-10
adds Dockerfile with janus and ustreamer installation

Esser50K:
Here is a Dockerfile with the step-by-step installation instructions including compiling the ustreamer plugin and running it with janus.

---
This change is [Reviewable](https://reviewable.io/reviews/tiny-pilot/ansible-role-tinypilot/171)


mtlynch:
Thanks! This isn't ready to go into `master`, so I created #172.

PR #173, opened 2022-01-12
Fix start.sh --h264-sink flag to match janus.plugin.ustreamer.jcfg

ottah:
This change is [Reviewable](https://reviewable.io/reviews/tiny-pilot/ansible-role-tinypilot/173)


Esser50K:
The change is correct ;)

PR #180, opened 2022-02-10
Increase CircleCI build instance to large

mtlynch:
There's also xlarge, which is 2x bigger, but in my experience, that actually results in slower builds because it takes 2-3 minutes to provision an xlarge machine.

---
This change is [Reviewable](https://reviewable.io/reviews/tiny-pilot/ansible-role-tinypilot/180)

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -25,6 +25,7 @@ jobs:
   build:
     machine:
       image: ubuntu-2004:202010-01
+    resource_class: large
     steps:
       - checkout
       - run:


jdeanwallace (approved):
:lgtm: 

> *[Reviewable](https://reviewable.io/reviews/tiny-pilot/ansible-role-tinypilot/180#-:-MvdNwoKCmlAjGqjT5ZC:bg7igkv)* status: :shipit: complete! 1 of 1 LGTMs obtained

PR #186, opened 2022-03-03
Refactor USB gadget library

jotaen4tinypilot:
These are the changes pulled out from https://github.com/tiny-pilot/ansible-role-tinypilot-pro/pull/86/files.

- No behavioural changes.
- The library file is apparently supposed to be the same in both repos (which I think makes sense). So I kept it that way, even if we don’t actually need `usb_gadget_deactivate` here.
- For all shell variables that I touched, [I added `{`…`}`](https://google.github.io/styleguide/shellguide.html#variable-expansion).

---
This change is [Reviewable](https://reviewable.io/reviews/tiny-pilot/ansible-role-tinypilot/186)

### DIFF
--- a/files/init-usb-gadget
+++ b/files/init-usb-gadget
@@ -127,16 +127,14 @@ echo -ne \\xC0           # END_COLLECTION
 } >> "$D"
 cp "$D" "${USB_MOUSE_FUNCTIONS_DIR}/report_desc"
 
-CONFIG_INDEX=1
-CONFIG_INDEX_DIR="${USB_CONFIGS_DIR}/c.${CONFIG_INDEX}"
-mkdir -p "$CONFIG_INDEX_DIR"
-echo 250 > "${CONFIG_INDEX_DIR}/MaxPower"
+mkdir -p "${USB_CONFIG_DIR}"
+echo 250 > "${USB_CONFIG_DIR}/MaxPower"
 
-CONFIGS_STRINGS_DIR="${CONFIG_INDEX_DIR}/${USB_STRINGS_DIR}"
-mkdir -p "$CONFIGS_STRINGS_DIR"
-echo "Config ${CONFIG_INDEX}: ECM network" > "${CONFIGS_STRINGS_DIR}/configuration"
+CONFIGS_STRINGS_DIR="${USB_CONFIG_DIR}/${USB_STRINGS_DIR}"
+mkdir -p "${CONFIGS_STRINGS_DIR}"
+echo "Config ${USB_CONFIG_INDEX}: ECM network" > "${CONFIGS_STRINGS_DIR}/configuration"
 
-ln -s "$USB_KEYBOARD_FUNCTIONS_DIR" "${CONFIG_INDEX_DIR}/"
-ln -s "$USB_MOUSE_FUNCTIONS_DIR" "${CONFIG_INDEX_DIR}/"
+ln -s "${USB_KEYBOARD_FUNCTIONS_DIR}" "${USB_CONFIG_DIR}/"
+ln -s "${USB_MOUSE_FUNCTIONS_DIR}" "${USB_CONFIG_DIR}/"
 
 usb_gadget_activate
--- a/files/lib/usb-gadget.sh
+++ b/files/lib/usb-gadget.sh
@@ -3,17 +3,20 @@
 export readonly USB_DEVICE_DIR="g1"
 export readonly USB_GADGET_PATH="/sys/kernel/config/usb_gadget"
 export readonly USB_DEVICE_PATH="${USB_GADGET_PATH}/${USB_DEVICE_DIR}"
-export readonly USB_FILE_CONFIG_PATH="${USB_DEVICE_PATH}/functions/mass_storage.0/lun.0/file"
-export readonly USB_CDROM_CONFIG_PATH="${USB_DEVICE_PATH}/functions/mass_storage.0/lun.0/cdrom"
 
 export readonly USB_STRINGS_DIR="strings/0x409"
 export readonly USB_KEYBOARD_FUNCTIONS_DIR="functions/hid.keyboard"
 export readonly USB_MOUSE_FUNCTIONS_DIR="functions/hid.mouse"
-export readonly USB_MASS_STORAGE_FUNCTIONS_DIR="functions/mass_storage.0"
+export readonly USB_MASS_STORAGE_NAME="mass_storage.0"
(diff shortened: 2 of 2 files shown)


jdeanwallace (commented):
> Reviewed 2 of 2 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/tiny-pilot/ansible-role-tinypilot/186)* status: 0 of 1 LGTMs obtained

___
*[files/init-usb-gadget, line 130 at r1](https://reviewable.io/reviews/tiny-pilot/ansible-role-tinypilot/186#-MxIVU-v1zo0YTUnmv3m:-MxIVU-v1zo0YTUnmv3n:bt78d3e) ([raw file](https://github.com/tiny-pilot/ansible-role-tinypilot/blob/c629a41acdbed56cb331b47cbbfca8e0dc6886e3/files/init-usb-gadget#L130)):*
> ```
> cp "$D" "${USB_MOUSE_FUNCTIONS_DIR}/report_desc"
> 
> mkdir -p "${USB_CONFIG_DIR}"
> ```

If we prefer absolute paths, can we use the new `USB_CONFIG_PATH` variable here (and everywhere else) and drop the `USB_CONFIG_DIR` variable from the lib?
___
*[files/lib/usb-gadget.sh, line 11 at r1](https://reviewable.io/reviews/tiny-pilot/ansible-role-tinypilot/186#-MxIThY82NB8MmYTei89:-MxIThY95yMl58a_2hSP:b-y0g1d4) ([raw file](https://github.com/tiny-pilot/ansible-role-tinypilot/blob/c629a41acdbed56cb331b47cbbfca8e0dc6886e3/files/lib/usb-gadget.sh#L11)):*
> ```Shell
> export readonly USB_MOUSE_FUNCTIONS_DIR="functions/hid.mouse"
> export readonly USB_MASS_STORAGE_NAME="mass_storage.0"
> export readonly USB_MASS_STORAGE_FUNCTIONS_PATH="${USB_DEVICE_PATH}/functions/${USB_MASS_STORAGE_NAME}"
> ```

(No action required) I like that we're moving to using absolute paths instead of relative paths. However, the inconsistency between the two bothers me :sweat_smile: But we'll probably align on a standard once we've established one.

One issue I have with the naming of the absolute paths is that it's less obvious now whether the path is pointing to a file or directory. To be fair, we haven't yet defined a variable that points to a file. So far, all absolute paths we've defined have been to directories. So this isn't really an issue if we assume all `_PATH` variables refer to directories.

jotaen4tinypilot (commented):
> *[Reviewable](https://reviewable.io/reviews/tiny-pilot/ansible-role-tinypilot/186)* status: 0 of 1 LGTMs obtained

___
*[files/init-usb-gadget, line 130 at r1](https://reviewable.io/reviews/tiny-pilot/ansible-role-tinypilot/186#-MxIVU-v1zo0YTUnmv3m:-MxInAdn3_J1Jo-dH4bi:bcqao4g) ([raw file](https://github.com/tiny-pilot/ansible-role-tinypilot/blob/c629a41acdbed56cb331b47cbbfca8e0dc6886e3/files/init-usb-gadget#L130)):*

If we prefer absolute paths, can we use the new `USB_CONFIG_PATH` variable here (and everywhere else) and drop the `USB_CONFIG_DIR` variable from the lib?
</blockquote></details>

I see your point about the consistency in the usage of absolute or relative paths.

When revisiting the other scripts, it seems like a common pattern for us (at least as far as USB gadget logic is concerned) is using relative paths and switching the working directory in the scripts.

I think if we we wanted it consistent, it’s maybe better to keep it all with relative paths in `lib/usb-gadget.sh` for now. (I’ve pushed a commit to change it to that.)

For https://github.com/tiny-pilot/ansible-role-tinypilot-pro/pull/86, we could then either do:

- Use `pushd`/`popd` to change the working directory of the script
- Keep using absolute paths, but either concatenate them in place, or have local helper variables.
___
*[files/lib/usb-gadget.sh, line 11 at r1](https://reviewable.io/reviews/tiny-pilot/ansible-role-tinypilot/186#-MxIThY82NB8MmYTei89:-MxIgE2Y2T1L5G38soBf:b-jwvpnt) ([raw file](https://github.com/tiny-pilot/ansible-role-tinypilot/blob/c629a41acdbed56cb331b47cbbfca8e0dc6886e3/files/lib/usb-gadget.sh#L11)):*

(No action required) I like that we're moving to using absolute paths instead of relative paths. However, the inconsistency between the two bothers me :sweat_smile: But we'll probably align on a standard once we've established one.

One issue I have with the naming of the absolute paths is that it's less obvious now whether the path is pointing to a file or directory. To be fair, we haven't yet defined a variable that points to a file. So far, all absolute paths we've defined have been to directories. So this isn't really an issue if we assume all `_PATH` variables refer to directories.
</blockquote></details>

We actually have two variables with absolute paths pointing to a file, but we are just about to remove them 😀 (`USB_FILE_CONFIG_PATH` and `USB_CDROM_CONFIG_PATH`).

jdeanwallace (approved):
I've also tested on device and works as usual :lgtm:

> Reviewed 1 of 1 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/tiny-pilot/ansible-role-tinypilot/186#-:-MxYUSI0EKaUqLshW9Td:b-ka83z0)* status: :shipit: complete! 1 of 1 LGTMs obtained

___
*[files/init-usb-gadget, line 130 at r1](https://reviewable.io/reviews/tiny-pilot/ansible-role-tinypilot/186#-MxIVU-v1zo0YTUnmv3m:-MxYNnqk8ZBM_exh0BZK:b-cp2719) ([raw file](https://github.com/tiny-pilot/ansible-role-tinypilot/blob/c629a41acdbed56cb331b47cbbfca8e0dc6886e3/files/init-usb-gadget#L130)):*
<details><summary><i>Previously, jotaen4tinypilot wrote…</i></summary><blockquote>

I see your point about the consistency in the usage of absolute or relative paths.

When revisiting the other scripts, it seems like a common pattern for us (at least as far as USB gadget logic is concerned) is using relative paths and switching the working directory in the scripts.

I think if we we wanted it consistent, it’s maybe better to keep it all with relative paths in `lib/usb-gadget.sh` for now. (I’ve pushed a commit to change it to that.)

For https://github.com/tiny-pilot/ansible-role-tinypilot-pro/pull/86, we could then either do:

- Use `pushd`/`popd` to change the working directory of the script
- Keep using absolute paths, but either concatenate them in place, or have local helper variables.
</blockquote></details>

Nice, yeah I think the consistently using relative paths looks better. Thanks.

PR #187, opened 2022-03-08
Simplify and re-order variables for USB gadget

jotaen4tinypilot:
Tiny code-cleanup in `lib/usb-gadget.sh`, without any behavioural effect:

We ended up with two variables that have almost the same name, `USB_CONFIGS_DIR` and `USB_CONFIG_DIR`. Apart from that causing confusion, we don’t actually need to export the former, since [the only usage is in the file itself](https://github.com/tiny-pilot/ansible-role-tinypilot/search?q=USB_CONFIGS_DIR). But for simplicity, we can as well just remove the variable altogether and repeat the value. We do the same already for the `functions/` substring in other variables, which I find readable and straightforward.

I’ve also grouped and reordered the variables, because it felt more sensible that way.

---
This change is [Reviewable](https://reviewable.io/reviews/tiny-pilot/ansible-role-tinypilot/187)

### DIFF
--- a/files/lib/usb-gadget.sh
+++ b/files/lib/usb-gadget.sh
@@ -10,12 +10,10 @@ export readonly USB_MOUSE_FUNCTIONS_DIR="functions/hid.mouse"
 export readonly USB_MASS_STORAGE_NAME="mass_storage.0"
 export readonly USB_MASS_STORAGE_FUNCTIONS_DIR="functions/${USB_MASS_STORAGE_NAME}"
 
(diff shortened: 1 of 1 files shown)


mtlynch (approved):
(taking over for Erwann, who's on vacation until Thursday)

:lgtm: 

> *[Reviewable](https://reviewable.io/reviews/tiny-pilot/ansible-role-tinypilot/187#-:-MxdKIi229A0gnN3b_-n:b-qt8fos)* status: :shipit: complete! 1 of 1 LGTMs obtained

Issue #168, opened 2021-12-28
Integrate Janus for WebRTC


mtlynch:
As part of [H264 integration](https://github.com/tiny-pilot/tinypilot/issues/875), we need to integrate a Janus server and configure it to work with TinyPilot's nginx configuration.

Ideally, we'll use the existing [bitsy-ai/ansible-role-janus-gateway](https://github.com/bitsy-ai/ansible-role-janus-gateway) Ansible role. Let me know if we are, and I'll create a local fork.

We'll also have to update the [target uStreamer version](https://github.com/tiny-pilot/ansible-role-tinypilot/blob/c565b6020c419db31d3fb94c86f5cf019c56908c/vars/main.yml#L13) to the latest, as we're on 3.26, which precedes Janus/H264 support.

Janus should be off by default with an Ansible variable that controls whether or not to install it.

### Blocked on

- [x] ~~ansible-role-tinypilot: [Research: Create a proof of concept Janus-uStreamer-TinyPilot integration](https://github.com/tiny-pilot/ansible-role-tinypilot/issues/167)~~
- [x] ~~ansible-role-ustreamer: [Add support for building Janus plugin](https://github.com/tiny-pilot/ansible-role-ustreamer/issues/49)~~

jdeanwallace:
> Ideally, we'll use the existing [bitsy-ai/ansible-role-janus-gateway](https://github.com/bitsy-ai/ansible-role-janus-gateway) Ansible role. Let me know if we are, and I'll create a local fork.

@mtlynch Can you make a local fork for me please?
We'll need to make some alterations to enforce idempotency and skip creating plugin settings we won't need.

PR #190, opened 2022-03-22
Tinypilot Janus integration

jdeanwallace:
This PR is part of https://github.com/tiny-pilot/ansible-role-tinypilot/issues/168

This PR installs the Janus Gateway on a TinyPilot device and enables H264 video streaming (without manual intervention).

This PR relies on https://github.com/tiny-pilot/ansible-role-janus-gateway/pull/1 to be merged.

### Changes
1. Added `ansible-role-janus-gateway` as a dependency ([line](https://github.com/tiny-pilot/ansible-role-tinypilot/blob/98134a2818fffd6f47ada9ea7a7aadcbbe2e8b92/requirements.yml#L4)).
1. Removed the Janus Docker folder and config files ([diff](https://github.com/tiny-pilot/ansible-role-tinypilot/commit/98134a2818fffd6f47ada9ea7a7aadcbbe2e8b92)).
1. Changed the Janus websocket port from `8002` -> `8188` because that's the one hardcoded in the [`ansible-role-janus-gateway`](https://github.com/tiny-pilot/ansible-role-janus-gateway/blob/a3b852ac6582fb139bbdb30c7b6cb5038cf6eaa5/templates/janus.transport.websockets.jcfg#L13) and the setting is not yet controlled by an ansible variable. The updated port works, so I didn't bother updating the role to manage this setting because we'll probably change the `ansible-role-janus-gateway` structure a bit anyway to suit our needs ([see comment here](https://github.com/tiny-pilot/ansible-role-janus-gateway/pull/1#issue-1176364322)).


mtlynch:
At this point, we can make this change in `master`. The install should be off by default, but we can start making these role changes in our real role.

PR #200, opened 2022-05-20
Install pre-built Janus Debian package

### LINKED ISSUE #199: Use pre-built Janus Debian package instead of building it from source
Janus takes 20-30m to build from source on a Raspberry Pi device. Now that we have [a pre-built Janus binary](https://github.com/tiny-pilot/janus-debian/releases/tag/1.0.1), we should install that instead of compiling it from scratch at install time.

jdeanwallace:
Resolves #199

This PR removes the dependency on [ansible-role-janus-gateway](https://github.com/tiny-pilot/ansible-role-janus-gateway) (which compiles Janus from source) and instead installs our [pre-built Janus Debian package](https://github.com/tiny-pilot/janus-debian/releases/tag/1.0.1).

I've tested the following upgrade paths:
1. TinyPilot, without Janus -> TinyPilot with Janus (pre-built)
2. TinyPilot with Janus (compiled from source) -> TinyPilot with Janus (pre-built)

With regards to situation 2., the pre-built Janus package needs to install [conflicting dependencies](https://github.com/tiny-pilot/janus-debian/blob/1.0.1/Dockerfile#L172) (that is shipped with the Debian package). Seeing as these dependencies are already installed, we need to [force the Janus installation](https://github.com/tiny-pilot/ansible-role-tinypilot/blob/1ace7f946858c25cabdfbb4f819685613751c8d5/tasks/janus.yml#L5-L8).  The [`dpkg` manual](https://man7.org/linux/man-pages/man1/dpkg.1.html) considers this dangerous, but seeing as we created this conflict to begin with, I think it should be okay 😬 
> **[force-]conflicts**: Install, even if it conflicts with another
              package. This is dangerous, for it will usually cause
              overwriting of some files.  This affects the Conflicts
              field.

Also related to situation 2., this PR doesn't attempt to remove any old files/config from the target device that has been left there from the previous Janus installation. Regardless of that, WebRTC streaming seems to work fine after the upgrade.

---
This change is [Reviewable](https://reviewable.io/reviews/tiny-pilot/ansible-role-tinypilot/200)

### DIFF
--- a/handlers/main.yml
+++ b/handlers/main.yml
@@ -20,3 +20,8 @@
   service:
     name: usb-gadget
     state: started
+
+- name: restart janus service
+  service:
+    name: janus
+    state: restarted
--- a/requirements.yml
+++ b/requirements.yml
@@ -1,4 +1,3 @@
 ---
 - src: https://github.com/tiny-pilot/ansible-role-ustreamer
 - src: https://github.com/tiny-pilot/ansible-role-nginx
-- src: https://github.com/tiny-pilot/ansible-role-janus-gateway
--- a/tasks/janus.yml
+++ b/tasks/janus.yml
@@ -2,7 +2,3 @@
 - name: install Janus Debian package
   apt:
     deb: "{{ janus_deb_file }}"
-    # We need to force the installation of Janus when the target device has
-    # conflicting dependencies. This does not install Janus on every provision.
-    # https://github.com/tiny-pilot/janus-debian/blob/1.0.1/Dockerfile#L172
-    force: yes
--- a/tasks/janus.yml
+++ b/tasks/janus.yml
@@ -2,3 +2,7 @@
 - name: install Janus Debian package
   apt:
     deb: "{{ janus_deb_file }}"
+    # We need to force the installation of Janus when the target device has
+    # conflicting dependencies. This does not install Janus on every provision.
+    # https://github.com/tiny-pilot/janus-debian/blob/1.0.1/Dockerfile#L172
+    force: yes
--- a/tasks/janus.yml
+++ b/tasks/janus.yml
@@ -1,4 +1,4 @@
 ---
-- name: import Janus Gateway role
-  import_role:
-    name: ansible-role-janus-gateway
+- name: install Janus Debian package
+  apt:
+    deb: "{{ janus_deb_file }}"
--- a/tasks/ustreamer.yml
+++ b/tasks/ustreamer.yml
@@ -1,11 +1,4 @@
 ---
-# We're unable to build the uStreamer Janus plugin because an include statement
-# references the wrong file path.
-# Issue: https://github.com/tiny-pilot/ansible-role-tinypilot/issues/192
-- name: patch Janus plugin.h file to successfully include refcount.h file
-  command: sed -i -e 's|^#include "refcount.h"$|#include "../refcount.h"|g' "{{ janus_install_dir }}/include/janus/plugins/plugin.h"
-  when: tinypilot_install_janus and ustreamer_compile_janus_plugin
-
 - name: import uStreamer role
   import_role:
     name: ansible-role-ustreamer
@@ -13,7 +6,9 @@
 - name: create uStreamer Janus plugin config
   template:
     src: janus.plugin.ustreamer.jcfg.j2
-    dest: "{{ janus_conf_dir }}/janus.plugin.ustreamer.jcfg"
+    dest: "{{ janus_install_dir }}/etc/janus/janus.plugin.ustreamer.jcfg"
+  notify:
+    - restart janus service
   when: >-
     tinypilot_install_janus
     and ustreamer_compile_janus_plugin
@@ -24,6 +19,8 @@
     remote_src: yes
     src: "{{ ustreamer_dir }}/janus/libjanus_ustreamer.so"
     dest: "{{ janus_install_dir }}/lib/janus/plugins/"
+  notify:
+    - restart janus service
   when: >-
     tinypilot_install_janus
     and ustreamer_compile_janus_plugin
--- a/vars/main.yml
+++ b/vars/main.yml
@@ -12,6 +12,9 @@ ustreamer_port: 8001
 ustreamer_repo: https://github.com/tiny-pilot/ustreamer.git
 ustreamer_repo_version: v4.13
 
-janus_version: v1.0.0
+janus_deb_file: https://github.com/tiny-pilot/janus-debian/releases/download/1.0.1/janus_1.0.1-20220519_armhf.deb
+# These variables are only used within this role and don't affect the Janus
+# installation config.
+janus_install_dir: /opt/janus
 janus_ws_ip: '127.0.0.1'
 janus_ws_port: 8002


mtlynch (changes requested):
> On the device, run the follow commands to enable H264 streaming:

Won't this install it from source with the Ansible role rather than with the new Debian package?

> *[Reviewable](https://reviewable.io/reviews/tiny-pilot/ansible-role-tinypilot/200#-:-N2q_Acc3kONwSMBgcH-:b-3apotn)* status: 0 of 1 LGTMs obtained

___
*[`requirements.yml` line 4 at r1](https://reviewable.io/reviews/tiny-pilot/ansible-role-tinypilot/200#-N2ljMmR2SL-Z78wXIlr:-N2qL1v271Wf--E2-98a:bh848jx) ([raw file](https://github.com/tiny-pilot/ansible-role-tinypilot/blob/63996cfd1d9cabb3a26d9c1a3dcea20ee3b24a55/requirements.yml#L4)):*

I'm reluctant to add cleanup code because it's an eyesore and it'll have to live in our codebase forever.

We don't have a great way of dealing with unused dependencies (even for our pip requirements), we just always add and never remove.

So leaving an unused `/opt/tinypilot-updater/ansible-role-janus-gateway` directory on disk doesn't seem like a big deal either.
</blockquote></details>

Yeah, I think leaving the old role is fine. I believe it's less than 1 MB of data, so I'm not too worried about leaving it behind, especially since it will only be on systems that tried the experimental opt-in.
___
*[`tasks/janus.yml` line 8 at r1](https://reviewable.io/reviews/tiny-pilot/ansible-role-tinypilot/200#-N2q_f-_-wgXwetFNevM:-N2q_f-_-wgXwetFNevN:b-cddbye) ([raw file](https://github.com/tiny-pilot/ansible-role-tinypilot/blob/1ace7f946858c25cabdfbb4f819685613751c8d5/tasks/janus.yml#L8)):*
> ```YAML
>     # conflicting dependencies. This does not install Janus on every provision.
>     # https://github.com/tiny-pilot/janus-debian/blob/1.0.1/Dockerfile#L172
>     force: yes
> ```

Is this just for people who installed the experimental version? If so, I'd rather just update our [install instructions](https://tinypilotkvm.com/blog/whats-new-in-2022-05#enabling-h264-video) to add `sudo apt-get remove -y libwebsockets ...` at the  top.

jdeanwallace (commented):
> Won't this install it from source with the Ansible role rather than with the new Debian package?

Yes, sorry I thought that's what @jotaen4tinypilot was looking for.

If not, this is how you can run re-provison a TinyPilot device using the current branch:

1. Install the latest version TinyPilot on a hobbyist device. You can do this by flashing an SD card with the latest build or to install it via the `quick-install` script.
2. On your dev machine, pull down the latest changes for this branch.
3. On your dev machine, follow the [Local Testing](https://github.com/tiny-pilot/ansible-role-tinypilot-pro#local-testing) instructions for a hobbyist device.
    * In order to enable H264 on the device, use the following command to generate the appropriate ansible playbook:
        ```bash
        TINYPILOT_PLAYBOOK=$(mktemp)
        echo "- hosts: all
          become: true
          become_method: sudo
          roles:
            - role: ${PWD}
              vars:
                ustreamer_h264_sink: tinypilot::ustreamer::h264
                ustreamer_h264_sink_mode: 777
                ustreamer_h264_sink_rm: yes
                ustreamer_compile_janus_plugin: yes
                tinypilot_install_janus: yes" > "${TINYPILOT_PLAYBOOK}"
        ```
4. Now you should have the Janus systemd service running and WebRTC enabled on the frontend.

> *[Reviewable](https://reviewable.io/reviews/tiny-pilot/ansible-role-tinypilot/200#-:-N2qbIAa7x-G66lMIMqk:btqg9lx)* status: 0 of 1 LGTMs obtained

jdeanwallace (commented):
> *[Reviewable](https://reviewable.io/reviews/tiny-pilot/ansible-role-tinypilot/200)* status: 0 of 1 LGTMs obtained

___
*[`tasks/janus.yml` line 8 at r1](https://reviewable.io/reviews/tiny-pilot/ansible-role-tinypilot/200#-N2q_f-_-wgXwetFNevM:-N2vH5yY--4KSrh0ZEUO:beydmza) ([raw file](https://github.com/tiny-pilot/ansible-role-tinypilot/blob/1ace7f946858c25cabdfbb4f819685613751c8d5/tasks/janus.yml#L8)):*
> Is this just for people who installed the experimental version?

Yes.

Done, thanks I've updated the install instructions here: https://github.com/tiny-pilot/tinypilotkvm.com/pull/734

jotaen4tinypilot (approved):
:lgtm: (Tested again on device.)

> Reviewed 2 of 5 files at r1, 1 of 1 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/tiny-pilot/ansible-role-tinypilot/200#-:-N2zt_rX9efiT8xVAIss:bs7n4e4)* status: :shipit: complete! 1 of 1 LGTMs obtained

___
*[`tasks/ustreamer.yml` line 16 at r2](https://reviewable.io/reviews/tiny-pilot/ansible-role-tinypilot/200#-N2ztnH70MUC6w0iE-_Y:-N2ztnH70MUC6w0iE-_Z:b6uxkus) ([raw file](https://github.com/tiny-pilot/ansible-role-tinypilot/blob/63996cfd1d9cabb3a26d9c1a3dcea20ee3b24a55/tasks/ustreamer.yml#L16)):*
> ```YAML
>   template:
>     src: janus.plugin.ustreamer.jcfg.j2
>     dest: "{{ janus_conf_dir }}/janus.plugin.ustreamer.jcfg"
> ```

Two question here, just for my own understanding:

- Had `janus_conf_dir` been defined anywhere previously? I.e., we don’t have to delete any definition now that we removed the last usage of the variable?
- `janus_install_dir` is [now defined by us explicitly](https://github.com/tiny-pilot/ansible-role-tinypilot/pull/200/files#diff-b8e57d5bc17b5679065938c3a8fd9be95bf3e5b1b6c80c1e4bee652492ae740fR18), but [we were already using it before](https://github.com/tiny-pilot/ansible-role-tinypilot/pull/200/files#diff-3e9ac118f46e7a365f2191a7417100e680dfa4bf9ba09a163f79593ef0b11a32L26). This is obviously not a problem, right?

Are these some sort of magic Ansible variables? ([I couldn’t find anything about it in the docs, though.](https://docs.ansible.com/ansible/latest/reference_appendices/special_variables.html))

jdeanwallace (commented):
> *[Reviewable](https://reviewable.io/reviews/tiny-pilot/ansible-role-tinypilot/200)* status: :shipit: complete! 1 of 1 LGTMs obtained

___
*[`tasks/ustreamer.yml` line 16 at r2](https://reviewable.io/reviews/tiny-pilot/ansible-role-tinypilot/200#-N2ztnH70MUC6w0iE-_Y:-N3-Blxg4Gl0nTA-CQxJ:b29si6a) ([raw file](https://github.com/tiny-pilot/ansible-role-tinypilot/blob/63996cfd1d9cabb3a26d9c1a3dcea20ee3b24a55/tasks/ustreamer.yml#L16)):*
> * Had `janus_conf_dir` been defined anywhere previously? I.e., we don’t have to delete any definition now that we removed the last usage of the variable?
> * `janus_install_dir` is now defined by us explicitly, but we were already using it before. This is obviously not a problem, right?

Both `janus_conf_dir` and `janus_install_dir` were defined in [`ansible-role-janus-gateway`](https://github.com/tiny-pilot/ansible-role-janus-gateway/blob/a22e9ef998b00ca65b1574c9266a0232d791b577/defaults/main.yml#L24-L25), so when we import the role we're also able to reference the variables it defines.

From the [docs for the `import_role` module](https://docs.ansible.com/ansible/latest/collections/ansible/builtin/import_role_module.html#notes):
> Since Ansible 2.7 variables defined in `vars` and `defaults` for the role are exposed to the play at playbook parsing time. Due to this, these variables will be accessible to roles and tasks executed before the location of the `ansible.builtin.import_role` task.

That's also why it's important to namespace variables (with a prefix like `tinypilot_`) because it's easy to have variable naming collisions between roles.

So to answer your questions, we need explicitly define `janus_install_dir` because we're no longer importing the `ansible-role-janus-gateway` role.

PR #202, opened 2022-06-09
Add an option for installing TinyPilot from a Debian package

### LINKED ISSUE #201: Update overhaul: provide non-git option to install `tinypilot` folder
In addition to our [current way of installing the `tinypilot` folder via git](https://github.com/tiny-pilot/ansible-role-tinypilot/blob/bd7f56a8b81744d03f376c96e6ce4623a8832248/tasks/main.yml#L104-L145), we need to offer an `--extra-vars` option so that the `tinypilot` folder can be created from a local Debian package. (The default way is still like it is right now, though.)

The new option could be `tinypilot_install_source`:
- If it’s `null`, then default to git
- If it’s a local path to a Debian package, use that package

[The Debian package will be bundled into the tarball](https://github.com/tiny-pilot/tinypilot-bundler/issues/6), so the new update mechanism is supposed to consume this option.

* [Scratch implementation](https://github.com/tiny-pilot/ansible-role-tinypilot/compare/experimental/with-bundler?expand=1)

mtlynch:
This adds an new variable `tinypilot_debian_package_path` that allows the user to install TinyPilot via a Debian package instead of from git.

Fixes #201

---

How I tested the role:

First, manually install `python3-venv` on a fresh Raspberry Pi OS install (`python3-venv` is not yet declared as a dependency in the prototype .deb).

Then, run this playbook from a machine with Ansible:

```yaml
---
- hosts: raspberrypi
  become: True
  become_user: root
  become_method: sudo
  roles:
    - role: ansible-role-tinypilot
      vars:
        tinypilot_debian_package_path: "https://p.tinypilotkvm.com/!TzCpZrSrjS/tinypilot-pro-2.4.1-1-armhf.deb"
```

We don't have a TinyPilot Community .deb package yet, so the above installs TinyPilot Pro on a TinyPilot Community system. The web app doesn't totally work because it's expecting HTTPS, but it places the files correctly, so it gives me confidence that it will work once we have the Community .deb file.

---
This change is [Reviewable](https://reviewable.io/reviews/tiny-pilot/ansible-role-tinypilot/202)

### DIFF
--- a/defaults/main.yml
+++ b/defaults/main.yml
@@ -2,6 +2,9 @@
 tinypilot_group: tinypilot
 tinypilot_dir: /opt/tinypilot
 tinypilot_privileged_dir: /opt/tinypilot-privileged
+# Specifies the path to a Debian package that installs TinyPilot. If unset, the
+# role installs TinyPilot from source using the tinypilot_repo variable instead.
+tinypilot_debian_package_path: null
 tinypilot_repo: https://github.com/tiny-pilot/tinypilot.git
 tinypilot_repo_branch: master
 # Port on which TinyPilot frontend listens (accessible from other hosts on the
--- a/defaults/main.yml
+++ b/defaults/main.yml
@@ -2,8 +2,9 @@
 tinypilot_group: tinypilot
 tinypilot_dir: /opt/tinypilot
 tinypilot_privileged_dir: /opt/tinypilot-privileged
-# Specifies the path to a Debian package that installs TinyPilot. If unset, the
-# role installs TinyPilot from source using the tinypilot_repo variable instead.
+# Specifies the filesystem path or URL of a Debian package that installs
+# TinyPilot. If unset, the role installs TinyPilot from source using the
+# tinypilot_repo variable instead.
 tinypilot_debian_package_path: null
 tinypilot_repo: https://github.com/tiny-pilot/tinypilot.git
 tinypilot_repo_branch: master
--- a/tasks/main.yml
+++ b/tasks/main.yml
@@ -55,7 +55,7 @@
   apt:
     name: "{{ tinypilot_packages }}"
     state: present
-  when: tinypilot_debian_package_path is null
+  when: tinypilot_debian_package_path is None
 
 - name: create tinypilot group
   group:
@@ -105,7 +105,7 @@
 - name: install TinyPilot Debian package
   apt:
     deb: "{{ tinypilot_debian_package_path }}"
-  when: tinypilot_debian_package_path is not null
+  when: tinypilot_debian_package_path
 
 - name: create TinyPilot folder
   file:
@@ -114,7 +114,7 @@
     owner: "{{ ansible_user }}"
     group: "{{ ansible_user }}"
     recurse: yes
-  when: tinypilot_debian_package_path is null
+  when: tinypilot_debian_package_path is None
   tags:
     # This fails the idempotency test because it must run on every provision
     # to set the correct directory ownership that allows git to sync.
@@ -130,7 +130,7 @@
     dest: "{{ tinypilot_dir }}"
     version: "{{ tinypilot_repo_branch }}"
     accept_hostkey: yes
-  when: tinypilot_debian_package_path is null
+  when: tinypilot_debian_package_path is None
   notify:
     - restart TinyPilot service
 
--- a/tasks/main.yml
+++ b/tasks/main.yml
@@ -55,7 +55,7 @@
   apt:
     name: "{{ tinypilot_packages }}"
     state: present
-  when: tinypilot_debian_package_path is none
+  when: tinypilot_debian_package_path == None
 
 - name: create tinypilot group
   group:
@@ -105,7 +105,7 @@
 - name: install TinyPilot Debian package
   apt:
     deb: "{{ tinypilot_debian_package_path }}"
-  when: tinypilot_debian_package_path is not none
+  when: tinypilot_debian_package_path != None
 
 - name: create TinyPilot folder
   file:
@@ -114,7 +114,7 @@
     owner: "{{ ansible_user }}"
     group: "{{ ansible_user }}"
     recurse: yes
-  when: tinypilot_debian_package_path is none
+  when: tinypilot_debian_package_path == None
   tags:
     # This fails the idempotency test because it must run on every provision
     # to set the correct directory ownership that allows git to sync.
@@ -130,7 +130,7 @@
     dest: "{{ tinypilot_dir }}"
     version: "{{ tinypilot_repo_branch }}"
     accept_hostkey: yes
-  when: tinypilot_debian_package_path is none
+  when: tinypilot_debian_package_path == None
   notify:
     - restart TinyPilot service
 
--- a/tasks/main.yml
+++ b/tasks/main.yml
@@ -55,7 +55,7 @@
   apt:
     name: "{{ tinypilot_packages }}"
     state: present
-  when: tinypilot_debian_package_path is None
+  when: tinypilot_debian_package_path is none
 
 - name: create tinypilot group
   group:
@@ -105,7 +105,7 @@
 - name: install TinyPilot Debian package
   apt:
     deb: "{{ tinypilot_debian_package_path }}"
-  when: tinypilot_debian_package_path
+  when: tinypilot_debian_package_path is not none
 
 - name: create TinyPilot folder
   file:
@@ -114,7 +114,7 @@
     owner: "{{ ansible_user }}"
     group: "{{ ansible_user }}"
     recurse: yes
-  when: tinypilot_debian_package_path is None
+  when: tinypilot_debian_package_path is none
   tags:
     # This fails the idempotency test because it must run on every provision
     # to set the correct directory ownership that allows git to sync.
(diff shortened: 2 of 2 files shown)


jdeanwallace (approved):
:lgtm: 

> Reviewed 2 of 2 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/tiny-pilot/ansible-role-tinypilot/202#-:-N4CZmN48Ug-TIiROJEh:bg7igkv)* status: :shipit: complete! 1 of 1 LGTMs obtained

___
*[`tasks/main.yml` line 58 at r1](https://reviewable.io/reviews/tiny-pilot/ansible-role-tinypilot/202#-N4CFE6C13dOcqafXW06:-N4CFE6C13dOcqafXW07:budj4yc) ([raw file](https://github.com/tiny-pilot/ansible-role-tinypilot/blob/79ca3bf2ff96bd3b61562e1a04a9eb95cbc1a7a7/tasks/main.yml#L58)):*
> ```YAML
>     name: "{{ tinypilot_packages }}"
>     state: present
>   when: tinypilot_debian_package_path is none
> ```

(no action required) I initial thought this lowercase none was wrong, but it's correct. It's not supposed to be Python, but rather a [Jinja test](https://jinja.palletsprojects.com/en/3.1.x/templates/#list-of-builtin-tests).

Alternatively (& confusingly), this would also work:
```yaml
when: tinypilot_debian_package_path == None
```
___
*[`tasks/main.yml` line 58 at r1](https://reviewable.io/reviews/tiny-pilot/ansible-role-tinypilot/202#-N4CPPjA0Zoq_PfHNd-q:-N4CPPjA0Zoq_PfHNd-r:bz93aji) ([raw file](https://github.com/tiny-pilot/ansible-role-tinypilot/blob/79ca3bf2ff96bd3b61562e1a04a9eb95cbc1a7a7/tasks/main.yml#L58)):*
> ```YAML
>     name: "{{ tinypilot_packages }}"
>     state: present
>   when: tinypilot_debian_package_path is none
> ```

Does this mean that the Debian package will be responsible for installing it's own dependencies?

If we skip this step, then [pip won't be installed](https://github.com/tiny-pilot/ansible-role-tinypilot/blob/79ca3bf2ff96bd3b61562e1a04a9eb95cbc1a7a7/tasks/main.yml#L44-L52) and may cause our [pip dependencies not be installed](https://github.com/tiny-pilot/ansible-role-tinypilot/blob/79ca3bf2ff96bd3b61562e1a04a9eb95cbc1a7a7/tasks/main.yml#L146-L153).

Suggestion: we haven't really defined what responsibilities our TinyPilot Debian package will have, so let's not be too strict here and let this step run for now.

mtlynch (commented):
> *[Reviewable](https://reviewable.io/reviews/tiny-pilot/ansible-role-tinypilot/202)* status: :shipit: complete! 1 of 1 LGTMs obtained

___
*[`tasks/main.yml` line 58 at r1](https://reviewable.io/reviews/tiny-pilot/ansible-role-tinypilot/202#-N4CFE6C13dOcqafXW06:-N4CdxKi028i0Ltce0Vy:b-ks4cvl) ([raw file](https://github.com/tiny-pilot/ansible-role-tinypilot/blob/79ca3bf2ff96bd3b61562e1a04a9eb95cbc1a7a7/tasks/main.yml#L58)):*

(no action required) I initial thought this lowercase none was wrong, but it's correct. It's not supposed to be Python, but rather a [Jinja test](https://jinja.palletsprojects.com/en/3.1.x/templates/#list-of-builtin-tests).

Alternatively (& confusingly), this would also work:
```yaml
when: tinypilot_debian_package_path == None
```
</blockquote></details>

Yeah, this confused me too. I remembered `is None` as working and went through a few build failures.

I like `== None` since it's not as confusing.
___
*[`tasks/main.yml` line 58 at r1](https://reviewable.io/reviews/tiny-pilot/ansible-role-tinypilot/202#-N4CPPjA0Zoq_PfHNd-q:-N4CdZms9CfuIo6kks2X:b-xmmg5f) ([raw file](https://github.com/tiny-pilot/ansible-role-tinypilot/blob/79ca3bf2ff96bd3b61562e1a04a9eb95cbc1a7a7/tasks/main.yml#L58)):*

Does this mean that the Debian package will be responsible for installing it's own dependencies?

If we skip this step, then [pip won't be installed](https://github.com/tiny-pilot/ansible-role-tinypilot/blob/79ca3bf2ff96bd3b61562e1a04a9eb95cbc1a7a7/tasks/main.yml#L44-L52) and may cause our [pip dependencies not be installed](https://github.com/tiny-pilot/ansible-role-tinypilot/blob/79ca3bf2ff96bd3b61562e1a04a9eb95cbc1a7a7/tasks/main.yml#L146-L153).

Suggestion: we haven't really defined what responsibilities our TinyPilot Debian package will have, so let's not be too strict here and let this step run for now.
</blockquote></details>

Wouldn't they get installed automatically if the Debian package declares its dependencies?

https://github.com/tiny-pilot/tinypilot-pro/blob/7e7ed0f4a735c68d33166996504bee97f428a163/Dockerfile#L34

The problem with leaving in the `apt` play is that it's really slow, and it installs `git`, which we hopefully won't need if we switch to Debian packages.

jdeanwallace (commented):
> *[Reviewable](https://reviewable.io/reviews/tiny-pilot/ansible-role-tinypilot/202)* status: :shipit: complete! 1 of 1 LGTMs obtained

___
*[`tasks/main.yml` line 58 at r1](https://reviewable.io/reviews/tiny-pilot/ansible-role-tinypilot/202#-N4CPPjA0Zoq_PfHNd-q:-N4Cgykm5hK4CV8XPHKQ:b-vkfpq) ([raw file](https://github.com/tiny-pilot/ansible-role-tinypilot/blob/79ca3bf2ff96bd3b61562e1a04a9eb95cbc1a7a7/tasks/main.yml#L58)):*
> Wouldn't they get installed automatically if the Debian package declares its dependencies?

Yeah, I think you're right. I was thinking about [a situation when using `dpkg`](https://github.com/tiny-pilot/janus-debian#install).

> The problem with leaving in the apt play is that it's really slow, and it installs git, which we hopefully won't need if we switch to Debian packages.

Then should we [directly install pip here instead](https://github.com/tiny-pilot/ansible-role-tinypilot/blob/2d2082c00572e988388a7a85113d6a198b1f9788/tasks/main.yml#L44-L52) of appending to the list of apt dependencies?

mtlynch (commented):
> *[Reviewable](https://reviewable.io/reviews/tiny-pilot/ansible-role-tinypilot/202)* status: :shipit: complete! 1 of 1 LGTMs obtained

___
*[`tasks/main.yml` line 58 at r1](https://reviewable.io/reviews/tiny-pilot/ansible-role-tinypilot/202#-N4CPPjA0Zoq_PfHNd-q:-N4CnI2898YuqL4Ah9kt:b-wkc56q) ([raw file](https://github.com/tiny-pilot/ansible-role-tinypilot/blob/79ca3bf2ff96bd3b61562e1a04a9eb95cbc1a7a7/tasks/main.yml#L58)):*

> Wouldn't they get installed automatically if the Debian package declares its dependencies?

Yeah, I think you're right. I was thinking about [a situation when using `dpkg`](https://github.com/tiny-pilot/janus-debian#install).

> The problem with leaving in the apt play is that it's really slow, and it installs git, which we hopefully won't need if we switch to Debian packages.

Then should we [directly install pip here instead](https://github.com/tiny-pilot/ansible-role-tinypilot/blob/2d2082c00572e988388a7a85113d6a198b1f9788/tasks/main.yml#L44-L52) of appending to the list of apt dependencies?
</blockquote></details>

Is there something unique about the pip package in particular that requires us to install it directly? I was thinking it would just be one of the apt dependencies we declare from the TinyPilot Debian package. Would that not work?

If it's about making sure we install python-pip3 vs python-pip, I don't mind just hardcoding the Debian package dependency to python3-pip and dealing with it later if someone needs to install on Debian <= 10 or Ubuntu <= 20.

mtlynch (commented):
> *[Reviewable](https://reviewable.io/reviews/tiny-pilot/ansible-role-tinypilot/202)* status: :shipit: complete! 1 of 1 LGTMs obtained

___
*[`tasks/main.yml` line 58 at r1](https://reviewable.io/reviews/tiny-pilot/ansible-role-tinypilot/202#-N4CPPjA0Zoq_PfHNd-q:-N4D-bRYAtHYdquRFm4F:b-ho6x1o) ([raw file](https://github.com/tiny-pilot/ansible-role-tinypilot/blob/79ca3bf2ff96bd3b61562e1a04a9eb95cbc1a7a7/tasks/main.yml#L58)):*

> Is there something unique about the pip package in particular that requires us to install it directly? I was thinking it would just be one of the apt dependencies we declare from the TinyPilot Debian package. Would that not work?

Oh okay, that makes sense. I was just fixating on pip because we use pip in the ansible and I wanted ansible to fulfill its own dependencies.
</blockquote></details>

Oh, that should be fine. If Ansible is running the play, it means that Ansible's dependencies are already in place. We bootstrap Ansible's dependencies in [quick-install](https://github.com/tiny-pilot/tinypilot/blob/2d82f7af60a1be3c99bb09cb7c2359a2abd0635e/quick-install#L134L164), so there's a separate process for ensuring pip is present by that point.

PR #205, opened 2022-06-29
Update update script to use new get-tinypilot.sh script

### LINKED ISSUE #204: Update Overhaul: Execute `get-tinypilot.sh` when performing a system update
The [control flow for executing a system update currently goes like this](https://github.com/tiny-pilot/tinypilot-pro/blob/experimental/update-overhaul/UPDATE-WORKFLOW.md#update-via-the-web-ui):

1. TinyPilot app triggers Updater systemd service
2. Updater systemd service executes [`scripts/update-service`](https://github.com/tiny-pilot/tinypilot-pro/blob/master/scripts/update-service)
3. `update-service` script calls [`update` privileged script](https://github.com/tiny-pilot/ansible-role-tinypilot-pro/blob/master/files/update)
4. `update` privileged script downloads and executes `quick-install` script
5. `update-service` writes result into JSON file, to communicate back to the TinyPilot app.

[The new flow that is described in the `UPDATE-WORKFLOW` document](https://github.com/tiny-pilot/tinypilot-pro/blob/experimental/update-overhaul/UPDATE-WORKFLOW.md#update-via-the-web-ui-1) doesn’t apply to Community: the situation in Community is simpler, since we always download the latest version, so we don’t have to save and pass down the desired version. Effectively, we need to change step 4:

1.-3. (just as before, see above)
4. **(new) `update` privileged script downloads `get-tinypilot.sh` script and executes it**
5. (just as before, see above)

Note: since the `quick-install` script needs to stay where it is for backwards-compatibility, we could also consider to always go through it, instead of changing the `update` privileged script. In this case, this issue would be obsolete. Pro: just one single way how it’s being done. Con: an unnecessary indirection just for legacy reasons. Note, the latter (the Con) will become more apparent in Pro, since we need to pass down additional parameters.

> **⚠️⚠️⚠️
> Please hold off on merging until the second 2.4.2 milestone is complete.
> (See [the note of the second milestone](https://github.com/tiny-pilot/tinypilot-pro/issues/445).)
> ⚠️⚠️⚠️**

mtlynch:
This is gated on merging the tinypilot update-overhaul branch into master

Related: https://github.com/tiny-pilot/tinypilot/pull/1038

Fixes #204

---
This change is [Reviewable](https://reviewable.io/reviews/tiny-pilot/ansible-role-tinypilot/205)

### DIFF
--- a/files/update
+++ b/files/update
@@ -12,5 +12,5 @@ set -e
 curl \
   --silent \
   --show-error \
-  https://raw.githubusercontent.com/tiny-pilot/tinypilot/master/quick-install | \
+  https://raw.githubusercontent.com/tiny-pilot/tinypilot/master/get-tinypilot.sh | \
     bash -
--- a/vars/main.yml
+++ b/vars/main.yml
@@ -1,7 +1,7 @@
 ---
-# TinyPilot's quick-install script (which it uses for installation and updates)
-# relies on the tinypilot user being named "tinypilot" so changing this value
-# will break updates.
+# TinyPilot's get-tinypilot.sh script (which it uses for installation and
+# updates) relies on the tinypilot user being named "tinypilot" so changing this
+# value will break updates.
 tinypilot_user: tinypilot
 
 # uStreamer variables are placed here, instead of in `defaults/main.yml`,


jdeanwallace:
A note on releasing the `update-overhaul`: This PR needs to merged just before we merge `update-overhaul` because the [bundle process packages up the master branch of this repo](https://github.com/tiny-pilot/tinypilot/blob/bbbf6493844aaa2445c621f54cf6710ae31b267c/bundler/create-bundle#L49-L60).

mtlynch:
Update: With #1071, the change-in-lockstep part is no longer required.

jdeanwallace:
@mtlynch - As part of https://github.com/tiny-pilot/tinypilot/pull/1076, I've created an [`update-overhaul` branch on `ansible-role-tinypilot`](https://github.com/tiny-pilot/ansible-role-tinypilot/tree/update-overhaul). So now it's safe to merge this PR with the `update-overhaul` branch instead of `master`.

PR #209, opened 2022-07-28
Update Nginx config for upstream reverse proxies

LukeLambert:
This is untested but might work. The `map` directives must be in the `http` block (outside of the `server` blocks). [See discussion](https://github.com/tiny-pilot/tinypilot/issues/1077).

<a data-ca-tag href="https://codeapprove.com/pr/tiny-pilot/ansible-role-tinypilot/209">Review on CodeApprove</a>

### DIFF
--- a/tasks/nginx.yml
+++ b/tasks/nginx.yml
@@ -11,6 +11,27 @@
   import_role:
     name: ansible-role-nginx
   vars:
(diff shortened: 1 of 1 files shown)


mtlynch:
Thanks for contributing! I've reopened the bug, and we'll revisit this when someone has time to test the fix.

PR #222, opened 2022-09-23
Add support for uStreamer 5.x on Bullseye

mtlynch:
This change adjusts the version of the uStreamer dependency depending on the version of Debian running on the target system.

For GPU-based encoding (omx in Debian Buster and earlier, m2m in Debian Bullseye and later), there's complicated compatibility:

|                       | Debian Buster (and earlier) | Debian Bullseye |
|---------------|-----------------------------|-----------------|
| uStreamer 4.x | Full compatibility | omx-based encoding doesn't work |
| uStreamer 5.x | m2m-based encoding doesn't work | Full compatibility |

This change configures TinyPilot's dependency on uStreamer based on the OS version running on the target system so that we install 4.x for a Buster-based system and 5.x for a Bullseye-based system.

Tested successfully on Buster and Bullseye with a TC358743 capture chip.

### DIFF
--- a/defaults/main.yml
+++ b/defaults/main.yml
@@ -26,3 +26,8 @@ tinypilot_app_settings_file: "/home/{{ tinypilot_user }}/app_settings.cfg"
 # (Experimental): Install the Janus WebRTC server to enable uStreamer to stream
 # video over H264 instead of MJPEG.
 tinypilot_install_janus: no
+
+# uStreamer version to use on systems prior to Raspbian Bullseye (Debian 11).
+ustreamer_repo_version_legacy: v4.13
+# uStreamer version to use on Raspbian Bullseye (Debian 11) or later systems.
+ustreamer_repo_version_modern: v5.22
--- a/defaults/main.yml
+++ b/defaults/main.yml
@@ -30,4 +30,4 @@ tinypilot_install_janus: no
 # uStreamer version to use on systems prior to Raspbian Bullseye (Debian 11).
 ustreamer_repo_version_legacy: v4.13
 # uStreamer version to use on Raspbian Bullseye (Debian 11) or later systems.
-ustreamer_repo_version_modern: v5.22
+ustreamer_repo_version_modern: v5.23
--- a/tasks/ustreamer.yml
+++ b/tasks/ustreamer.yml
@@ -6,11 +6,14 @@
 - name: choose the default version of uStreamer to install
   set_fact:
     ustreamer_repo_version: "{{ ustreamer_repo_version_modern }}"
+  when: ustreamer_repo_version is not defined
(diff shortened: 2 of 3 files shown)


mtlynch:
##### Automated comment from [CodeApprove ➜](https://codeapprove.com/pr/tiny-pilot/ansible-role-tinypilot/222)

⏳ @jdeanwallace please review this Pull Request

jdeanwallace (changes requested):
##### Automated comment from [CodeApprove ➜](https://codeapprove.com/pr/tiny-pilot/ansible-role-tinypilot/222)


Approved on CodeApprove
✔️ Approved (3 unresolved comments)

<hr />

In: **Discussion**
I haven't tested this on device, but I'm curious to know how you tested this on device now that our update overhaul makes that less fun?

<hr />

In: **tasks/ustreamer.yml**:

```
> Line 8
    ustreamer_repo_version: "{{ ustreamer_repo_version_modern }}"
```
We always override the `ustreamer_repo_version` which means that any version specified via `ustreamer_repo_version` (as extra playbook vars) will not be honored. Is that intended?

<hr />

In: **tasks/ustreamer.yml**:

```
> Line 19
    ustreamer_repo_version: "{{ ustreamer_repo_version }}"
```
I'm not 100% sure, but I think we can omit setting this role variable here because `ansible-role-ustreamer` will pick up the `ustreamer_repo_version` value that was set via the previous `set_fact` task 🤔 ([source](https://docs.ansible.com/ansible/latest/user_guide/playbooks_variables.html#understanding-variable-precedence))


<hr />

👀 @mtlynch it's your turn please take a look

mtlynch (commented):
##### Automated comment from [CodeApprove ➜](https://codeapprove.com/pr/tiny-pilot/ansible-role-tinypilot/222)



<hr />

In: **Discussion**
It's the same process as before since we still allow git-based installation of tinypilot in the Ansible role. You can't install on top of an existing new-style install (it will complain that /opt/tinypilot is not a git repo), but you can install fine on a clean Raspbian OS.

<hr />

In: **tasks/ustreamer.yml**:
Ah, you're right. Removed.

<hr />

In: **tasks/ustreamer.yml**:

```
> Line 8
    ustreamer_repo_version: "{{ ustreamer_repo_version_modern }}"
```
There aren't really cases where I'd expect the user to override, but maybe it will be useful for development to have that functionality, so I added conditions.

jdeanwallace (approved):
##### Automated comment from [CodeApprove ➜](https://codeapprove.com/pr/tiny-pilot/ansible-role-tinypilot/222)

Approved: I have approved this change on CodeApprove and all of my comments have been resolved.

PR #231, opened 2022-10-17
Retrieve the uStreamer version from the binary instead of git

mtlynch:
Versions of git with the 2022-04-12 security fix applied (https://github.blog/2022-04-12-git-security-vulnerability-announced/) will fail to check status if they're not run as the directory owner.

As a fix, we should just stop depending on git for the version number and use the output of ustreamer --version instead. This is also more flexible in the future if we decide to move from a source-compiled version to a precompiled Debian package for uStreamer.

Fixes https://github.com/tiny-pilot/tinypilot/issues/1140
<a data-ca-tag href="https://codeapprove.com/pr/tiny-pilot/ansible-role-tinypilot/231">Review on CodeApprove</a>

### DIFF
--- a/files/collect-debug-logs
+++ b/files/collect-debug-logs
@@ -71,7 +71,7 @@ cd /opt/tinypilot && \
 
 print_info "Checking uStreamer version..."
 cd /opt/ustreamer && \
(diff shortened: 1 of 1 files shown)


mtlynch:
##### Automated comment from [CodeApprove ➜](https://codeapprove.com/pr/tiny-pilot/ansible-role-tinypilot/231)

PR #233, opened 2022-10-19
Remove most tinypilot-privileged scripts from Ansible

mtlynch:
This change gets rid of scripts and sudoers logic that allow the `tinypilot` user to execute certain scripts as root. We are moving those scripts and sudoers logic to the core TinyPilot repo: https://github.com/tiny-pilot/tinypilot/pull/1148

We move `tinypilot_dir` and `tinypilot_privileged_dir` from `defaults` to `vars` because they're no longer values that an Ansible playbook is free to override, as the core TinyPilot repo assumes a directory of `/opt/tinypilot` and a privileged directory of `/opt/tinypilot-privileged`.


tasks/main.yml
```diff
@@ -7,21 +7,6 @@
...
-    src: "{{ item }}"
-    dest: "{{ tinypilot_privileged_dir }}/lib/{{ item | basename }}"
-    owner: root
-    group: root
-    mode: '0755'
-  with_fileglob:
-    - lib/*
-
 - name: install HID USB gadget
   import_tasks: install_usb_gadget.yml
```
jdeanwallace, line 11: There's a circular dependency here.

Some `install_usb_gadget.yml` tasks rely on the systemd `usb-gadget` service to be installed:
https://github.com/tiny-pilot/ansible-role-tinypilot/blob/3f2c12deb5c1fb56dee399dbcb234e2c941e3a9b/tasks/install_usb_gadget.yml#L60-L64
https://github.com/tiny-pilot/ansible-role-tinypilot/blob/3f2c12deb5c1fb56dee399dbcb234e2c941e3a9b/tasks/install_usb_gadget.yml#L78-L82
but now it only gets installed later via the TinyPilot Debian package:
https://github.com/tiny-pilot/ansible-role-tinypilot/blob/3f2c12deb5c1fb56dee399dbcb234e2c941e3a9b/tasks/main.yml#L51-L54
However, the usb-gadget can't be initialized without enabling the dwc2 driver to the boot config:
https://github.com/tiny-pilot/ansible-role-tinypilot/blob/3f2c12deb5c1fb56dee399dbcb234e2c941e3a9b/tasks/install_usb_gadget.yml#L30-L36

jdeanwallace (changes requested):

tasks/main.yml
mtlynch: Oh, good catch!

I think the best solution is to leave the usb-gadget stuff in the Ansible role for now and just focus on pulling out the other scripts. I'll update it but I might not have bandwidth until next week with the holidays.

mtlynch: @jdeanwallace - I've reverted the usb-gadget changes, so now the changes are only to files that the TinyPilot web app executes as root rather than any changes to systemd services.

jdeanwallace (approved):
LGTM

Issue #224, opened 2022-09-29
Install Janus by default (for WebRTC/H264 streaming)


jotaen4tinypilot:
In order for us to offer an H264 option in the web UI, we need to install Janus by default.

So the [default settings](https://github.com/tiny-pilot/ansible-role-tinypilot/blob/dbb7ef75b529bb8bb2dae7d14054dcd76a129b66/defaults/main.yml#L28) need to be changed:

```
tinypilot_install_janus: yes
```

We also need to add

```
ustreamer_compile_janus_plugin: yes
```

to make [ustreamer compile the janus plugin](https://github.com/tiny-pilot/ansible-role-ustreamer/blob/150b5de75968c880a566dd1bb8aeb0b8e7c12355/defaults/main.yml#L81-L82) right away.

To me, it would make sense to already setup the H264 configuration, otherwise we have to do that from the backend:

```
ustreamer_h264_sink: tinypilot::ustreamer::h264
ustreamer_h264_sink_mode: 777
ustreamer_h264_sink_rm: yes
```

We then have to remove the part where we [turn on WebRTC by default](https://github.com/tiny-pilot/ansible-role-tinypilot/blob/82566c663762a20d431a61cb8a07bce450799642/templates/tinypilot-app-settings.cfg.j2#L6-L9), if the respective configuration is present. Turning H264 on and off will be implemented in the web server.

PR #234, opened 2022-10-26
Install Janus (and therefore activate H264) by default

jotaen4tinypilot:
Resolves https://github.com/tiny-pilot/ansible-role-tinypilot/issues/224 by installing Janus by default on the system.

Note, retracting `USE_WEBRTC_REMOTE_SCREEN` from the app settings will disable WebRTC on users’ devices. They will manually have to turn it on again via the video settings dialog, since we [changed the mechanism how the streaming mode preference is saved](https://github.com/tiny-pilot/tinypilot/pull/1152). I’ve discussed this with @mtlynch, and as far as I remember, we said it’s okay that we make beta-users carry out that extra step to “re-opt-in” for H264.

## Molecule error

This PR is still draft, because the molecule is broken. We need to override the `ustreamer_h264_sink` variable in the molecule context, because otherwise the uStreamer role wants to install Janus Debian package, but that fails with error `Wrong architecture 'armhf'`. @mtlynch do you have an idea what the best way would be to solve this by any chance?
<a data-ca-tag href="https://codeapprove.com/pr/tiny-pilot/ansible-role-tinypilot/234">Review on CodeApprove</a>

### DIFF
--- a/files/update-video-settings
+++ b/files/update-video-settings
@@ -72,5 +72,6 @@ echo "- hosts: localhost
 ansible-playbook \
   "${PLAYBOOK}" \
   --inventory localhost, \
+  --extra-vars "@${INSTALLER_DIR}/roles/ansible-role-tinypilot/vars/main.yml" \
   --extra-vars "${EXTRA_VARS_PATH}" \
   --tags "systemd-config"


mtlynch (approved):
##### Automated comment from [CodeApprove ➜](https://codeapprove.com/pr/tiny-pilot/ansible-role-tinypilot/234)


<a href="https://codeapprove.com/pr/tiny-pilot/ansible-role-tinypilot/234">Approved on CodeApprove</a>
✔️ Approved

LGTM, thanks!

<hr />

👀 @jotaen4tinypilot it's your turn please take a look

Issue #237, opened 2022-10-31
Privileged `update-video-settings` script erroneously disregards installation variables


jotaen4tinypilot:
On a fresh install, `ansible-role-tinypilot` creates the following systemd file for the uStreamer service:

```
[Unit]
Description=uStreamer - Lightweight, optimized video encoder
After=syslog.target network.target

[Service]
Type=simple
User=ustreamer
WorkingDirectory=/opt/ustreamer
ExecStart=/opt/ustreamer/ustreamer \
  --host 127.0.0.1 \
  --port 8001 \
  --encoder omx \
  --format uyvy \
  --workers 3 \
  --drop-same-frames 30 \
  --persistent \
  --dv-timings \
  && :
# This last line is just to end the multi-line command because the line
# before is ending with backslash and so expects to be continued.
Restart=always

[Install]
WantedBy=multi-user.target
```

After running the privileged `update-video-settings` script (e.g. when applying new video settings from the UI), the `--host` flag goes missing:

```
[Unit]
Description=uStreamer - Lightweight, optimized video encoder
After=syslog.target network.target

[Service]
Type=simple
User=ustreamer
WorkingDirectory=/opt/ustreamer
ExecStart=/opt/ustreamer/ustreamer \
  --port 8001 \
  --encoder omx \
  --format uyvy \
  --workers 3 \
  --drop-same-frames 30 \
  --persistent \
  --dv-timings \
  && :
# This last line is just to end the multi-line command because the line
# before is ending with backslash and so expects to be continued.
Restart=always

[Install]
WantedBy=multi-user.target
```

The reason is that when the uStreamer role is initially run during the installation, we run it from within the context of `ansible-role-tinypilot`. So all variables from [`vars/main.yml`](https://github.com/tiny-pilot/ansible-role-tinypilot/blob/b6c905c4cf6ffd0f066a798ab0e93d6f2aa9261e/vars/main.yml) are read and set. However, when running the privileged `update-video-settings` script, [we run `ansible-role-ustreamer` directly](https://github.com/tiny-pilot/ansible-role-tinypilot/blob/b6c905c4cf6ffd0f066a798ab0e93d6f2aa9261e/files/update-video-settings#L62-L76), i.e. independent from the `ansible-role-tinypilot` context, so none of the parameters from `vars/main.yml` are present.

[...]

PR #238, opened 2022-10-31
Provide complete vars to uStreamer role when updating video settings

jotaen4tinypilot:
Resolves https://github.com/tiny-pilot/ansible-role-tinypilot/issues/237 by additionally passing on the parameters from `vars/main.yml` to the uStreamer role, when updating the video settings.

I’m not sure this is the cleanest solution to the problem, but to me it would seem good enough for now. An alternative approach could be to not execute `ansible-role-ustreamer` directly, but to have a task in `ansible-role-tinypilot` that’s like “update video settings”. That way we wouldn’t have to manually recreate the runtime environment, which obviously is prone to subtle bugs. But that might have other implications, such as degraded performance.

Note: the second `--extra-vars` seems to take precedence over the former. I [couldn’t find that stated in the Ansible docs](https://docs.ansible.com/ansible/latest/user_guide/playbooks_variables.html#defining-variables-at-runtime), so I just tried it out.
<a data-ca-tag href="https://codeapprove.com/pr/tiny-pilot/ansible-role-tinypilot/238">Review on CodeApprove</a>

### DIFF
--- a/files/update-video-settings
+++ b/files/update-video-settings
@@ -72,5 +72,6 @@ echo "- hosts: localhost
 ansible-playbook \
   "${PLAYBOOK}" \
   --inventory localhost, \
+  --extra-vars "@${INSTALLER_DIR}/roles/ansible-role-tinypilot/vars/main.yml" \
   --extra-vars "${EXTRA_VARS_PATH}" \
   --tags "systemd-config"


mtlynch (approved):
##### Automated comment from [CodeApprove ➜](https://codeapprove.com/pr/tiny-pilot/ansible-role-tinypilot/238)


<a href="https://codeapprove.com/pr/tiny-pilot/ansible-role-tinypilot/238">Approved on CodeApprove</a>
✔️ Approved

Thanks for catching this!

Yeah, this isn't ideal, but I agree it's kind of the best we can do for the moment.

<hr />

👀 @jotaen4tinypilot it's your turn please take a look

PR #244, opened 2022-12-13
Remove legacy nginx vhosts cleanup.

jdeanwallace:
Resolves https://github.com/tiny-pilot/ansible-role-tinypilot-pro/issues/132

This PR depends on first resolving https://github.com/tiny-pilot/gatekeeper/issues/83

Now that TinyPilot Community/Pro users are forced to upgrade to `1.8.0`/`2.5.0` (before upgrading any further), we can remove legacy cleanup code.

The `ansible-role-tinypilot-pro` version of this PR (i.e., merging community changes) would be to remove [this ansible task](https://github.com/tiny-pilot/ansible-role-tinypilot-pro/blob/1ce12942a5ef711f5481d8600a321c6c54f43cc4/tasks/nginx.yml#L11-L18).
<a data-ca-tag href="https://codeapprove.com/pr/tiny-pilot/ansible-role-tinypilot/244">Review on CodeApprove</a>

### DIFF
--- a/tasks/nginx.yml
+++ b/tasks/nginx.yml
@@ -1,12 +1,4 @@
 ---
-# TinyPilot Pro has two nginx configuration files, whereas TinyPilot Community
-# has just one. Remove the extra Pro file otherwise nginx will fail due to the
-# extra file if the user downgrades from Pro to free.
-- name: remove stray conf file from TinyPilot Pro
-  file:
-    path: /etc/nginx/sites-enabled/tinypilot.http.conf
-    state: absent
-
 - name: import nginx role
   import_role:
     name: ansible-role-nginx


jdeanwallace:
@jotaen4tinypilot - The review of this PR is low priority because it depends on the issue you're working on https://github.com/tiny-pilot/gatekeeper/issues/83

mtlynch:
This is different than the line we're talking about in https://github.com/tiny-pilot/ansible-role-tinypilot-pro/issues/132.

The Community Ansible role still has to get rid of the Pro config if the user does a Pro -> Community downgrade, but the Pro role no longer has to purge the config file that hasn't existed for two years.

jdeanwallace:
@mtlynch - I think it is the same legacy config file. In both repos, the legacy file being removed is named `tinypilot.http.conf`:
* Pro: https://github.com/tiny-pilot/ansible-role-tinypilot-pro/blob/1ce12942a5ef711f5481d8600a321c6c54f43cc4/tasks/nginx.yml#L15
* Community: https://github.com/tiny-pilot/ansible-role-tinypilot/blob/07220b946873d4f49253765d309de02eeb5daf32/tasks/nginx.yml#L7

And in both repos the main nginx config being used is called `tinypilot.conf`:
* Pro: https://github.com/tiny-pilot/ansible-role-tinypilot-pro/blob/1ce12942a5ef711f5481d8600a321c6c54f43cc4/tasks/nginx.yml#L37-L38
* Community: https://github.com/tiny-pilot/ansible-role-tinypilot/blob/07220b946873d4f49253765d309de02eeb5daf32/tasks/nginx.yml#L29
    (results in the file being named `tinypilot.conf`)

So in the case of a Pro -> Community downgrade, the main nginx config file will be overwritten.

mtlynch:
@jdeanwallace - Oh, sorry you're right. I thought we were still cleaning up the old `kvmpi.conf` from back when TinyPilot was (briefly) called "KVM Pi." I forgot that we also unified HTTP + HTTPS into a single config.

mtlynch (approved):
##### Automated comment from [CodeApprove ➜](https://codeapprove.com/pr/tiny-pilot/ansible-role-tinypilot/244)


<a href="https://codeapprove.com/pr/tiny-pilot/ansible-role-tinypilot/244">Approved on CodeApprove</a>
✔️ Approved

LGTM

<hr />

👀 @jdeanwallace, @jotaen4tinypilot it's your turn please take a look

mtlynch:
@jotaen4tinypilot - Ah, right. I forgot which one was the blocker. Reverting in #245.

PR #197, opened 2022-04-28
Change directory owner before syncing git

jdeanwallace:
Part of https://github.com/tiny-pilot/tinypilot/issues/963

Depends on https://github.com/tiny-pilot/ansible-role-ustreamer/pull/59

Just echoing the description of above related PR:
> This is needed to allow git to sync after the announced [git security fix](https://github.blog/2022-04-12-git-security-vulnerability-announced/).
> 
> > By default, Git will refuse to even parse a Git config of a repository owned by someone else
> 
> Alternatively, we could bypass this new git behaviour by marking the git repo directory as a [`safe.directory`](https://git-scm.com/docs/git-config/2.35.2#Documentation/git-config.txt-safedirectory), but that would defeat the purpose of the security fix especially seeing as we can't assume the device is a single-user machine.

---
This change is [Reviewable](https://reviewable.io/reviews/tiny-pilot/ansible-role-tinypilot/197)

PR #248, opened 2022-12-22
Add a workaround for git permissions restrictions

mtlynch:
In #197, we adjusted the TinyPilot directory's permissions to work around a git security fix:

> This is needed to allow git to sync after the announced [git security fix](https://github.blog/2022-04-12-git-security-vulnerability-announced/).
> 
> > By default, Git will refuse to even parse a Git config of a repository owned by someone else

We were conservative about overriding the security fix at the time because we still used git to install TinyPilot. Now, we only use the git-based install for testing, so we can be more aggressive about disabling the security setting.

I'm not sure why the previous fix suddenly stopped working, but it's not worth investigating at this point given that this flow is only used for testing.

<a data-ca-tag href="https://codeapprove.com/pr/tiny-pilot/ansible-role-tinypilot/248">Review on CodeApprove</a>

### DIFF
--- a/tasks/main.yml
+++ b/tasks/main.yml
@@ -72,8 +72,6 @@
   file:
     path: "{{ tinypilot_dir }}"
     state: directory
-    owner: "{{ ansible_user }}"
-    group: "{{ ansible_user }}"
     recurse: yes
   when: tinypilot_debian_package_path == None
   tags:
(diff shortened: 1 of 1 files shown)


mtlynch:
##### Automated comment from [CodeApprove ➜](https://codeapprove.com/pr/tiny-pilot/ansible-role-tinypilot/248)

mtlynch:
Closing, as #251 makes this obsolete

PR #252, opened 2022-12-23
Skip installing tinypilot user and group

mtlynch:
It's faster/simpler to create the `tinypilot` user and group in the Debian package as opposed to the Ansible role.

https://github.com/tiny-pilot/tinypilot/pull/1240 makes the corresponding change to the TinyPilot Debian package.

The ansible role still references the `tinypilot_group` variable, so we can't remove it entirely yet, but this change moves it to `vars` to hint to clients that it's not a variable they should change.

<a data-ca-tag href="https://codeapprove.com/pr/tiny-pilot/ansible-role-tinypilot/252">Review on CodeApprove</a>

### DIFF
--- a/README.md
+++ b/README.md
@@ -10,7 +10,6 @@ Ansible role for [TinyPilot KVM](https://github.com/tiny-pilot/tinypilot).
 Available variables are listed below, along with default values (see [defaults/main.yml](defaults/main.yml)):
 
 ```yaml
-tinypilot_group: tinypilot
 tinypilot_interface: '127.0.0.1'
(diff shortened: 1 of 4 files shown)


mtlynch:
##### Automated comment from [CodeApprove ➜](https://codeapprove.com/pr/tiny-pilot/ansible-role-tinypilot/252)

mtlynch:
@jotaen4tinypilot - CI is passing now that https://github.com/tiny-pilot/tinypilot/pull/1240 is merged

